Issue #40, opened 2024-07-31
[Pulverizer] Output acts like normal inventory slots, not like output slots.


jshipley:
### Issue description

This may affect other machines, but I found it on the Pulverizer.

Generally, you should be able to pull items out of an output slot, but not put items into an output slot. The Pulverizer allows you to put any item into the output slot.

As an example, if you have 54 raw copper in the input slot, and shift-click on a stack of 30 raw copper in your inventory, it should move 10 raw copper to the input slot. Instead, it moves 10 raw copper to the input slot and 20 raw copper to the output slot. 

Adding the excess raw copper to the output slot is not an expected or desirable behavior. Restricting input to the input slot (like a vanilla furnace) would be a better behavior.

### Minecraft version

1.21 (Latest)

### Fabric loader version

0.16.0

### Fabric API version

0.100.8+1.21

### Oritech version

0.8.0

Rearth:
Thanks for the suggestion (and also the PR). That's my favority kind of issue, with a PR right along it. i merged your PR, so this will be fixed with 0.8.1

The4codeblocks:
well I liked the way I was able to put the small dusts there and have them automatically merge without the hassle of crafting them

PR #45, opened 2024-08-01
Make machine output slots output-only

jshipley:
For #40 

Added an output parameter to the GuiSlot record, which makes the slot output only. Also went through and updated the machines to make them use output slots instead of regular slots.


Rearth:
This looks great, thanks for the PR. I'll just go ahead and merge this. It'll be included in update 0.8.1 then

Issue #42, opened 2024-07-31
Changing input item in Pulverizer doesn't reset progress


jshipley:
### Issue description

If you have (for example) raw copper being pulverized, and replace it with coal when the progress arrow is almost finished, the coal will start out almost finished instead of resetting the progress

### Steps to reproduce

1. Put raw copper in pulverizer
2. Wait until it's nearly done being pulverized
3. Replace raw copper with coal

Near instant pulverized coal :)

### Minecraft version

1.21 (Latest)

### Fabric loader version

0.16.0

### Fabric API version

0.100.8+1.21

### Oritech version

0.8.0

Rearth:
Thanks again for the PR for this. It's merged and fixed, so the 0.8.1 release will include this bugfix

PR #46, opened 2024-08-01
#42 - reset progress when changing input items

jshipley:
Added a simple check to reset progress if the candidate recipe is not the current recipe.


Rearth:
This is a great, simple solution. Thanks for contributing, I'll just go ahead and merge this

Rearth (approved):

Issue #38, opened 2024-07-30
Industrial door Issues


SpLatOniQ:
### Issue description

when breaking or placing a block next to the door the door makes an opening sound, the door's opening animation is too long for a button (not nessesarily an issue but its just annoying)

https://github.com/user-attachments/assets/b651fead-0611-448b-8897-bed32339c55d

### Minecraft version

1.21 (Latest)

### Fabric loader version

0.16.0

### Fabric API version

0.100.8+1.21

### Oritech version

0.8.0

Rearth:
This has been fixed thanks to @jshipley . The fix will be included in version 0.8.1

PR #47, opened 2024-08-01
#38 - only update door when powered state changes

jshipley:
The industrial door was making opening sounds any time there was a neighbor update.

This fixes it so that the sounds (and state changes) only happen when the powered status changes.

Also changed the industrial button tick duration from 50 to 80. That's a really long tick for a button, but it works a lot better with the industrial door animation.

As for note in #38 that regular buttons are too short for the industrial door, that seems like a choice. If you really want to use a regular button with the industrial door, you could make it work with more redstone.


Rearth (approved):

Rearth:
Very simple but effective fix, huge thanks for the PR. I'll go ahead and also merge it

PR #53, opened 2024-08-02
New texture, sprite sheet format, and updated models for item pipe.

jshipley:
The item pipes are retextured to look more like chests (the texture is similar, but simplified). The corners are machine orange, and the connector plate is also retextured.

I added arrow-like shapes to the sides of the connector plates. They sort of look a little like hoppers to me, though, which also seems fitting.


Rearth:
This looks great, thank you very much for the PR. I'll go ahead and merge it, so it's included in the next release

Issue #55, opened 2024-08-03
Duping with portable tank


jules552:
### Issue description

You can dupe any item using a portable tank that hold any amount of liquid in it.

### Steps to reproduce

1. Fill a portable tank with a liquid
2. Put an item into it
3. Quit the GUI 
4. Break the portable tank
5. Place the portable tank back, the item is still in it but as also been dropped

### Minecraft version

1.21 (Latest)

### Fabric loader version

0.16.0

### Fabric API version

0.100.8

### Oritech version

0.8.0

Rearth:
Thanks for reporting this issue, I completely missed it when implementing the custom tank block drops. I just fixed it with this commit: https://github.com/Rearth/Oritech/commit/8ce278bfa3708ab43d0b98b998ba9fd595f2d201
The bugfix will be included in the next release (0.8.1)

PR #56, opened 2024-08-03
Fixed duping issue with portable tank

jules552:
This PR address issue #55.

The `stack` variable had one issue since it kept containing the item that the block was holding inside his inventory but at the same time we were also dropping to the ground this same item, causing the duplication. 

I don't really know if the fix I've made is ideal, maybe you want to keep the functionnality of `writeNbt` the same but remove from the nbt data the item directly in the `onBreak` function?

### DIFF
--- a/src/main/java/rearth/oritech/block/blocks/machines/storage/SmallFluidTank.java
+++ b/src/main/java/rearth/oritech/block/blocks/machines/storage/SmallFluidTank.java
@@ -70,18 +70,29 @@ public class SmallFluidTank extends Block implements BlockEntityProvider {
     public BlockState onBreak(World world, BlockPos pos, BlockState state, PlayerEntity player) {
         
         if (!world.isClient) {
-            var stack = getStackWithData(world, pos);
             var tankEntity = (SmallFluidTankEntity) world.getBlockEntity(pos);
-            
-            if (!player.isCreative())
-                Block.dropStack(world, pos, stack);
-            
+
             var stacks = tankEntity.inventory.heldStacks;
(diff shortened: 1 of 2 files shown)


jshipley:
Wouldn't this make any buckets in the portable tank disappear when you log out?

A better fix would be to empty out the inventory slots when spawning items in the block's onBreak method.

jules552:
> Wouldn't this make any buckets in the portable tank disappear when you log out?
> 
> A better fix would be to empty out the inventory slots when spawning items in the block's onBreak method.

Yeah you were right it wasn't saving properly anymore, I've changed the way of handleling things so that everything is done in the `onBreak` method.
Tell me if it seems better to you. (Sorry I haven't really worked with Fabric in my life so)

Rearth:
Hi, thanks for the bugreport, and the PR for it. I see a few small issues with the changes here: Code from both the writeNbt and the itemstack creation is partially duplicated. I'm not sure if you saw it, but there's a method in the tankentity class for creating the itemstack with the proper data. This also includes a custom name, which breaks with your PR. I think the easiest and cleanest solution is to just remove the stored items from the dropped item nbt. This will work with any future changes to the nbt data, and requires way fewer changes.

I hope you don't mind it, but I just implemented this change on main while skipping this PR. As it's only one line, I think this way is easier than updating the PR to reflect the change. You can check it out here in this commit: https://github.com/Rearth/Oritech/commit/8ce278bfa3708ab43d0b98b998ba9fd595f2d201

Still, huge thanks for creating the PR. Even if I won't take this one, all help is appreciated.

PR #59, opened 2024-08-05
Remember extract state when switching between regular pipe and connector

jshipley:
Store "extract" state for Item and Fluid pipes when they are normal pipes.

This makes it so that a pipe that was a connector set to extract will still be set to extract any time it connects to another block.

For example, if a pipe is set to extract from a shulker box, then the shulker box is replaced by another shulker box, the pipe will still be set to extract from the new shulker box.


Rearth:
Merged, huge thanks for this. Never really thought about it before, but this is a great addition

PR #60, opened 2024-08-08
Portable storage piston/dispenser behaviors

jshipley:
With this change, portable tank and portable energy storage can be broken by pistons and placed by dispensers, just like shulker boxes. This enables redstone filling mechanisms that can keep replacing full storage blocks with empty storage blocks (or vice versa). Super useful for carrying large amounts of lava around (or letting drones carry them around).

The biggest change here is adding block_loot datagen for the Portable Tank and Portable Energy Storage that stores Nbt data in the dropped items.

onBreak() wasn't being called when the blocks were destroyed by pistons, so getDroppedStacks() was implemented to make sure that the blocks were dropping their inventory items when they are broken. This should work when broken by a player, or by a piston, or by an explosion.

Between the new block loot and getDroppedStacks(), the onBreak() methods were redundant and were removed.

### DIFF
--- a/src/main/generated/data/oritech/loot_table/blocks/small_storage_block.json
+++ b/src/main/generated/data/oritech/loot_table/blocks/small_storage_block.json
@@ -13,13 +13,6 @@
           "type": "minecraft:item",
           "functions": [
             {
-              "function": "minecraft:copy_components",
-              "include": [
-                "minecraft:custom_name"
-              ],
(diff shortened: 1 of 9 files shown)


Rearth:
Huh, didnt even know datapacks can do that. I'll keep this in mind for a few things in the future. Huge thanks again for the PR

PR #61, opened 2024-08-09
Make pipes connect to addons when addons are activated

jshipley:
Two small changes to addon initialization:

1) When settings an addon state to ADDON_USED, set the controller position before updating the block state.

When the block state changes, the neighbor pipes are updated. If the controller position isn't set first, then the pipes won't connect to the addon unless something else changes to make them update again.

This fixes the problem where pipes need to be broken and replaced to make them connect to an addon after it is activated.

2) When initializing a multiblock machine, also initialize the addons. This takes one step out of the initialization process.

If you're setting up a bunch of machines, you can add the machines, the machine cores, and the addons, then activate everything at once.

Or, if you have a machine with addons set up and you accidentally break a machine core you can replace the core and reactivate the machine.

This is a convenience, not a bugfix. You can already do this by right-clicking on the machine to initialize it, then right-clicking on the machine again to initialize the addons and show the menu.


Rearth:
Thanks again. I tried the 1. change with the pipes already, but didnt figure out that the order matters here

PR #63, opened 2024-08-09
Fix some z-fighting for lava generator, basic generator, and steam addon

jshipley:
Slightly deflated some block shapes to fix z-fighting issues.


Rearth:
Great improvement, those were also on my todo list. Really appreciate it

PR #64, opened 2024-08-09
Add silicon item tag for better compatibility with Energized Power

jshipley:
Energized Power is the other main tech pack available on 1.21 Fabric.

Adding c:silicon item tag so that Oritech silicon is interchangeable with EP silicon in recipes (EP already has silicon tagged as c:silicon).


Rearth:
Never really played much with energized power, I always though tech reborn is the only main other fabric tech mod. But thanks for the additinal, I'll also check out energized power

PR #65, opened 2024-08-09
Stop working when amethyst is fully grown.

jshipley:
Found a simple solution for continuous amethyst shard harvesting.

Stop firing laser at budding amethyst block when no more amethyst bud growth is possible.


Rearth:
Quite a simple solution, that's always good. It's an additional 6 blocks state checks per tick per laser, but I guess that's acceptable

jshipley:
Would it be better to do something like have a 20 tick (or more) delay between checks?

PR #67, opened 2024-08-16
Fixing inventory issues in portable tank and generator

jshipley:
This fixes a few inventory issues
* Portable Tank was getting in a broken state when using item pipes to move buckets (Items.EMPTY vs Items.AIR comparison)
* Portable Tank was letting item pipes extract from the input slot
* Basic Generator was not leaving bucket behind when using lava bucket as fuel

As a minor benefit, the portable tank bucket processing and generator now save nbt data for the buckets when emptying/filling them.


Rearth:
Great chanes, thanks

PR #70, opened 2024-08-18
Changed world gen to try to get a slightly more natural shape

jshipley:
…for resource nodes, oil wells, and ore boulders.

If you like the change, you may still want to play around with the generation parameters before merging it in.

** Resource Node **
Resource nodes are the biggest change, instead of carving out a square room with a flat layer of bedrock / resource nodes at the bottom, the world gen is carving out a bowl shape with resource nodes randomly scattered in the bedrock layers. Oil is still above that, and a thin pocket of air is added to the top. The whole resource node also had some noise added to make it not look like a perfect sphere.

The bedrock drill was also changed to work if it can find any resource nodes to drill underneath it (instead of checking if it was directly above a resource node), and keeps searching downward if the block it finds directly below is air. Particles will still display on all discovered drillable ores, so it should be obvious to the player what is going on and why some resource nodes may be unavailable. Players will also have to mine tuff/deepslate out of the way to potentially discover more resource nodes and get better performance out of the bedrock drill.

** Oil Spring **
It looked like it was *supposed* to generate with a shell of stone around the oil, so I fixed it so that it did. Also added some noise so that it doesn't look quite so much like a perfect sphere.

** Ore Boulder **

Only changed slightly. The random placement is slightly different (still only being moved up to radius - 1 blocks away which should prevent it from floating in the air or being completely covered in dirt), and some noise was added to the boulder shape. 

----

Even with the noise, the shapes are still mostly spherical. There would need to be some more work put into using better generation functions to get anything else. The sizes are all pretty small, though, and the contents look good in a mostly spherical shape, so I don't think that making them non-spherical would even matter much.

### DIFF
--- a/src/main/java/rearth/oritech/block/entity/machines/interaction/DeepDrillEntity.java
+++ b/src/main/java/rearth/oritech/block/entity/machines/interaction/DeepDrillEntity.java
@@ -5,6 +5,7 @@ import net.fabricmc.fabric.api.transfer.v1.item.ItemVariant;
 import net.fabricmc.fabric.api.transfer.v1.transaction.Transaction;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
+import net.minecraft.block.Blocks;
 import net.minecraft.block.entity.BlockEntity;
 import net.minecraft.block.entity.BlockEntityTicker;
 import net.minecraft.inventory.Inventories;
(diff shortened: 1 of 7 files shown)


Rearth:
I'll check this out in the coming days, thanks for the PR in advance already.

Rearth:
Sorry for the late reply, but I just tested it and these changes are looking great. Huge thanks

PR #71, opened 2024-08-18
Fixes #69, don't re-run onUseBlock for client

### LINKED ISSUE #69: Using offhand item when shift-clicking with target designator
### Issue description

When I shift-click with a target designator with torches in my offhand, the designator sets the target, but then I also place a torch with the same click. 

### Steps to reproduce

_No response_

### Minecraft version

1.21 (Latest)

### Fabric loader version

0.16.0

### Fabric API version

0.100.8+1.21

### Oritech version

0.8.1

### Other relevant versions

_No response_

_No response_

jshipley:
Running super.onUseBlock on the client side was causing the offhand action to occur AND the main action to occur.

Now, offhand actions will only occur if none of the potential main hand actions are handled.


Rearth:
Never really thought about offhand actions, but yeah this certainly was an issue. Huge thanks

PR #73, opened 2024-08-20
Adding compat recipes for Alloy Forgery

jshipley:
Added some recipes so that Alloy Forgery can make Oritech alloys (including alloying the gems to multiply ingots).

The recipes have a load condition so that they'll only load if Alloy Forgery is installed.

I opted to add json recipes directly instead of setting up datagen for these, because I didn't want to pull in Alloy Forgery as a build dependency or fork the Alloy Forgery recipe / serializer for ~ 12 recipes.


Rearth:
Thanks, using json recipes directly here makes sense for now. If we end up with a lot more recipes adding the extra datagen would make sense, but for this small additional I agree with just using json files.

PR #76, opened 2024-08-20
Tweaks for some machines

jshipley:
* Solar panel now only functions in sky light (some shade OK)
* Destroyer Block attempts to break block with shears if there was no drop loot
* Fertilizer can create grass/flowers on grass, or seagrass underwater
* Fertilizer can use bonemeal in gui slot to double fertilization
* Fertilizer should actually use energy now when fertilizing
* Pump block tests for infinite fluid instead of just treating all still water as infinite
* Tree Cutter stops working when inventory is full
* Tree Cutter treats warp blocks / glow shrooms as "leaves"
* Tree Cutter can also cut large mushrooms

Let me know if you want me to separate any of these out.


jshipley:
I just had a thought that the tree cutter could also cut chorus plants, especially because it would be somewhat likely to cut the chorus flowers first.. but then I thought that it might be stretching the "tree cutter" idea a little far. What do you think? It would just be adding chorus plants / fruits to the definition of "logs" and "leaves".

jshipley:
Another thing I just thought of (while working on the manual), should the block placer be able to use buckets to place fluid?

Rearth:
Hi, i really like most of those changes. For the wood cutters new block types, maybe we could add new block tags of things that the wood cutter can treat as tree? This could include the tags for the wood types, mushrooms, warped stems, and anything else pack makers might want to include?

Rearth:
Also I'm not too sure how I feel about the fertilizer additions. Is fertilizing underwater and fertilizing grass really a usecase that players have? Cause it adds a lot of complexity both code-wise and performance wise. I guess in most cases the change will be too small to notice a dip in performance, but on large servers with players spamming these farms it could be relevant. I want to keep the machines as simple as possible while still giving them as much function as needed

Rearth:
Same for the fluid infinity checks. It does this check on every block it mines, not just when initializing and then stopping the search early. The only real usecase for this I can see is players with the infinite lava gamerule enabled, which is very few I assume. I'm fine with treating a single water block as infinite, as plenty of mods add infinite water methods, and having this check for each pump does add up. Also it makes the logic a lot more complicated, while adding very little benefit to most uses

jshipley:
The tag idea for the wood cutter is way better, and would also make it easy for pack makers to include modded trees or other strange tree-like things. 

For the underwater fertilizing, I was mostly thinking about seagrass. That's a pretty niche farm, though, and I don't even know how often people want to breed turtles so much that they want to automate farming the seagrass. I was just thinking through all of the things that bonemeal can fertilize, and thinking that the fertilizer should be able to do it to. It's probably unnecessary. Same with the grass, it's useful for farming flowers for dye or bees. I can see how switching to bonemeal can be a lot more expensive, though.

Maybe for those, I'll revert the changes but keep doubling the fertilizer's growth (with the current method of increasing a crop's growth stage) if there's bonemeal (or other tagged fertilizer) in the fertilizer's inventory slot.

Would you be OK with the pump change if I made sure that it only did the check when initially caching pump positions? If not I can undo it and just let it pump water infinitely. As you said, that's going to be right most of the time. I'm not even aware of other infinite fluids in mods.

Thanks for entertaining my ideas, and as usual I really appreciate how much thought you've put into performance.

jshipley:
OK, changes are pushed in.

* The fertilizer work was reverted, except for c:fertilizers doubling the effectiveness of the machine.
* Pump block changes were reverted.
* Treecutter works with oritech:mineable/cutter_logs and oritech:mineable/cutter_leaves (let me know if you can think of a better tag name)

One change from before: mushroom stem is a "log" and red/brown mushroom blocks are "leaves" (instead of treating any mushroom block as a "log".

Also, Mangrove roots are a "log" here, but conceptually would it make more sense to add it to "leaves"?

Rearth:
Hi, the changes are looking good so far, I'll go ahead and merge it. Huge thanks again for the PR. Mangrove roots as logs should work just fine. The only different when handling leaves is the tree stem detection, shich shoudn't really matter here

PR #77, opened 2024-08-21
Added hitboxes to machine addons

jshipley:
* Added static initialization to machine addons for hitboxes with all facing/face rotations
  - Addon machine blocks that don't have a hitbox will default to full cube
* Added Geometry.rotateVoxelShape() utility function
* Converted machine addon parameter list to an AddonSettings record (with with* functions)
* Updated item blockstates / models to all have consistent rotation, lighting, item frame view, etc.


Rearth:
Damn that's a huge list of changes, but surely a welcome one. I really appreciate all the effort that went into this. Updates collision shapes are something that the addons definitly needed. Also didn't even know we coul;d adjust the lighting in the previews

PR #79, opened 2024-08-22
Guidebook art and additions

jshipley:
* Added "Getting Started" guide
* New "tablet" art (https://github.com/Rearth/Oritech/discussions/12) and sound effects - the book open sound effect is just an amethyst environmental sound. I don't hate it, but I also didn't find a sound elsewhere that I liked better.
* A lot of new pages for resources, machines, and tools
* Reorganized the content with subfolders for categories
* Fixed the lighting on some item models that were appearing very dark in the manual
* Replaced some multi-block renders with images created with isorender mod
* Added several images
* Added more links between pages
* Some changes to existing pages, mostly minor grammar fixes.

I'm not sure if I'll be able to read the manual for a while, but it'll probably need updating again before the 1.0 release to keep up with the changes.

Also, if anybody can do a better icon for the tablet, feel free :)


Rearth:
Huge thanks for this, the new guide book skin looks amazing. Not too sure if we want to explicitly mention recipes in the guidebook, as this may result in some issues with modpacks if they add their custom recipes. But then they could also override the relevant pages with datapacks, so I guess this should be fine

Rearth (approved):

PR #81, opened 2024-08-28
Make Oritech more translatable

jshipley:
I replaced Text.literal() and Text.of() with Text.translatable() everywhere I could.

I've done some testing and things are still looking good to me, but this affects messages, menus, tooltips, etc. all over the place.

In theory, everything should still look the same as it did before this change :)


OakPlatinum:
Hello! I am translating Oritech into Chinese. Should I use the localization file in this pull request or in main branch? Thank you.

jshipley:
The localization file in this branch adds new localization strings, and even then mostly just adds strings that were already literal strings in the main branch.

I'd recommend translating this file. It should be merged in before the next release.

Rearth:
Wow, this must have taken quite some effort. Thanks for the improvements. I just submitted some merge conflics and will proceed and merge it. Thank you again

Issue #43, opened 2024-07-31
Display multiblock in Jade (or Waila) instead of machine core


jshipley:
### Describe the feature you'd like

With all of the multiblock machines in Oritech, it would be nice if Jade could display any of the blocks in the machine as if it was the machine instead of displaying the original "machine block" when looking at part of the multiblock machine.

![2024-07-30_22 50 58](https://github.com/user-attachments/assets/cad3f650-745e-47a0-b061-bd2591c46948)

jshipley:
#82 (included in v0.9.0) shows both the machine core and the underlying machine in the Jade view.

PR #82, opened 2024-08-29
New Jade plugin to show machine core controllers

jshipley:
Here's a very basic Jade plugin.

It should add the machine controller's name in italics underneath the machine core name when looking at a machine core.

It turns out that just registering a server data provider for the machine core ended up automatically adding the fluid, energy, and storage contents to the Jade panel as well due to the way the providers were already implemented on the machine cores.

So now Jade should give all of the same information for machines and connected machine cores, while still making it very obvious that it's a machine core.

jade_machine
jade_core

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,6 +46,14 @@ repositories {
         url "https://modmaven.dev"
     }
 
+    maven { // for Jade
+        name = 'Curse'
+        url = 'https://www.cursemaven.com'
+        content {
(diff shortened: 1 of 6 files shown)


jshipley:
This implements https://github.com/Rearth/Oritech/issues/43

Rearth:
Huh, I had no idea that jade could even be extended like this. Huge thanks.

PR #83, opened 2024-08-29
Added some block variants for plated blocks

jshipley:
Added slabs, stairs, and pressure plates for all the "plated" blocks.

I thought about adding trapdoors as well, but trapdoors with good looking cutouts are beyond the level of simple datagen. At the very least, they'd need full models and interior textures so that they don't become x-ray devices.


Rearth:
Cool idea, thanks

PR #84, opened 2024-08-29
Machine hitboxes

jshipley:
Added hitboxes for the following:
* Block placer
* Fertilizer
* Item Filter
* Tree Cutter
* Basic Generator
* Pulverizer

I skipped the pump, but that can still be done before the next release if the new pump model isn't ready yet.


Rearth:
Hm, not sure if I like this addition yet. Adding and maintaining tight hitboxes seems like a lot of work for little advantage. Is there a tool in blockbench to create those automatically?
Also we'll have consistency issues with the multiblocks. For the addons it really made sense to add them as they are very small blocks. However, most machines are already somewhat close to a full block in size, and if we add tight hitboxes the multiblocks will feel out of line. Adding tight hitboxes to the multiblocks will probably be difficult.
If we find a way to fix this for the multiblocks aswell I'll really like this. Proper hitboxes indeed give the mod a much more finished feel

jshipley:
There is an addon in blockbench to generate voxelshapes for blocks, but it's not entirely compatible with the geo models. I had to convert the geo models into minecraft block models and then generate the shapes for them.

After conversion, some of the models were not positioned correctly, and others had pieces that were not angled/placed correctly, or were animated and I removed the moving parts from the voxel shapes. It still didn't take much work overall, especially since the voxel rotating code was already in place from the addons.

Multiblock is a whole different thing, though, because the machine cores are hidden but still breakable. It would probably be possible to give them empty hitboxes (making them all but non-existent when the machine is unpacked) and then create a new voxel shape for the entire machine.

As a player, I would love to have tighter hitboxes on the multiblock machines. The fuel generator feels especially weird when you're standing on top of it because the top of the model is about 1/4 block below the top of the hitbox. I'm also constantly 

I'll try changing the hitbox on a multi-block machine to see how easy it is, how the interaction feels, and whether it unnecessarily complicates the block logic. It might be a couple of weeks before I try it out, though.

Honestly, though, I'm not sure how a hitbox on something like the Enderic Laser would even work. The whole thing moves around so much that you'd have to dynamically generate a hitbox based on the block's animation state, give it some sort of static partial block hitbox that changes based on the facing but isn't really accurate to what the player sees, or just make it an exception and give it a full cube hitbox.

jshipley:
I'll reopen this when/if I get hitboxes on the multiblock machines.

PR #85, opened 2024-09-02
Enderic laser can now attack living entities

jshipley:
The core block searching / caching aspects of the enderic laser were not changed.

* Added new "Machine Hunter Addon" that lets the Enderic Laser target living entities instead of blocks
    - Damage done by laser depends on number of hunter addons (2^addon count, max 2^3)
    - Creepers hit by the laser will become charged
    - Crop Filter addon will make laser avoid baby animals (and in block destroy mode will avoid immature crops)
    - Quarry addon will affect hunting range (4^addon count, max 4^3)
    - Yield addon will function as Looting enchantment (max Looting 3)
    - Mobs killed by laser will not drop xp, items dropped will go into laser inventory (and any that don't fit will be automatically discarded)

* Added new "oritech:block/laser_passthrough" tag to define the blocks that the laser will ignore
* Using c:block/budding_blocks and c:block/buds for recognizing/"energizing" amethyst blocks to support other budding geodes (such as Geodes+), but fluxite harvesting is still specific to amethyst.

* Added behavior registry (separate registries for block and entity behaviors) for Enderic Laser (sort of like DispenserBehavior)
    - Makes it easy to add special handling for more block / entity types (such as cooling fusion reactors)
    - I think it's easier to follow, and it's definitely easier for addon mods to extend.
    
Note: for killing entities, the entity damage, damage sources, drops (loot, equipment, inventory), etc. is all complicated enough that I didn't want to try to reproduce it/simulate it in the laser class, so the laser just damages the living entities and lets them handle the health tracking, death, loot, etc, and then there's a simple mixin that checks if the killing entity is the laser and then gives the drops to the laser instead of dropping them if it is.

I also turned of XP drops when the laser does the killing, although it might be interesting to actually collect that XP somehow and make it usable for the player.


Rearth:
Hi, thanks for the PR. Great idea, this is something I was also planning on adding. Spreading the logic out into different files is a good idea. I just created a PR to your branch for some additional improvements: https://github.com/jshipley/Oritech/pull/1
as I didn't want to include all the changes via github PR suggestions. I do have a few more points before merging this:
- The model looks a bit too generic. Maybe instead of the small "knob" on the side we could have something that looks "aggressive", like a red orb or something
- I'm not too happy with the addon usage yet. The yield addons for looting is greeat. However, having the damage based on the hunter addon count and the range based on the quarry addon count feels a bit weird, when the laser is just damaging by basically pushing energy into the entity. Speed addons already increase the energy usage of the laser. Maybe we could scale the damage also by the amount of speed addons. So the more speed addons we have, the more energy is being used and that means more damaged inflicted? And then the hunter addons increase the scanning range? And then the UI also calls it scan range to make this clearer? Just as an idea
- Currently all living entities are being targeted. This also includes villagers and all friendly entities. Maybe we need some kind of mechanism to set it to hostiles only?

jshipley:
I've merged your changes back into my branch, so we can use that as a starting point now.

For the model, I actually tried to use a sword shape at first, but I wasn't happy with how it was looking. The current model is the texture from a magma block, as is the recipe. I'd love to have something better than that, though. I'll try out the red orb idea and see if I can make that work. I'd probably change the recipe to use something more suitable to the new design then, instead of the magma block.

Good ideas on the addon usage. I'll make the changes.

Yes, it currently does attack any non-player entities, which means that it would also attack pets and mounts. That's probably not a surprise that you'd want to spring on a player :)

There are a couple of ways that could be improved, but the best way would probably be to set the default to only attack hostile entities (but still target peaceful/player for beneficial interactions like charging chest armor). Then the target designator could be used to change the target mode, and an on-screen message would be sent to the player.

jshipley:
Something like this (but with better colors, I think the base needs to be a little darker, and the top should be bright, but a little less pink)? 
image

Rearth:
The new model is already looking much better. Using the target designator to switch hostile filtering sounds good

jshipley:
The "hunter" laser has three modes now:

* hostile - only target monsters
* hostile and neutral - target monsters, animals (unless they're pets), and water creatures
* all - target any non-player mobs (players will still be targeted for chestplate charging)

None of these will ever target a player for killing, and all will still target a player for charging an energy storage chestplate.

Also changed the addons as discussed, and added a new config option for base damage per tick.

src/main/resources/assets/oritech/lavender/entries/oritech_guide/interaction/enderic_laser.md
```diff
@@ -54,6 +54,21 @@ The laser ignores all input and output limits, and can fill the energy storage o
...
+**Machine Hunter Addon**
+
+When the enderic laser has a machine hunter addon, it will target mobs instead of blocks. Addon upgrades still work when hunting mobs: the yield addon will make the
+mobs drop more loot, the more hunter addons will extend the target scan range, the crop filter will avoid killing baby animals.
+
+If you are wearing an [exo chestplate](^oritech:tools/exo_armor) then the enderic laser will charge it for you.
+
+;;;;;
+
+** Hunter Targeting **
```
Rearth: The spaces cause invalid rendering

src/main/resources/assets/oritech/lavender/entries/oritech_guide/interaction/enderic_laser.md
```diff
@@ -54,6 +54,21 @@ The laser ignores all input and output limits, and can fill the energy storage o
 
 ;;;;;
 
+**Machine Hunter Addon**
+
+When the enderic laser has a machine hunter addon, it will target mobs instead of blocks. Addon upgrades still work when hunting mobs: the yield addon will make the
+mobs drop more loot, the more hunter addons will extend the target scan range, the crop filter will avoid killing baby animals.
```
Rearth: Maybe also mentioning that more energy transferred (e.g. speed addons) will increase the damage dealt

src/main/java/rearth/oritech/block/entity/machines/interaction/LaserArmBlockEntity.java
```diff
@@ -95,112 +129,107 @@ public void markDirty() {
...
     
     public LaserArmBlockEntity(BlockPos pos, BlockState state) {
         super(BlockEntitiesContent.LASER_ARM_ENTITY, pos, state);
+        laserHead = Vec3d.of(pos.up()).add(0.5, 0.55, 0.5);
     }
     
     @Override
     public void tick(World world, BlockPos pos, BlockState state, LaserArmBlockEntity blockEntity) {
-        if (world.isClient() || !isActive(state) || redstonePowered || currentTarget == null || energyStorage.getAmount() < energyRequiredToFire())
+        if (world.isClient() || !isActive(state) || redstonePowered || (hunterAddons == 0 && currentTarget == null) || energyStorage.getAmount() < energyRequiredToFire())
```
Rearth: I've noticed that the laser arm UI doesnt update the current energy amount when it's not active. E.g. when it's placed and charging up. This is probably caused by this section here

src/main/java/rearth/oritech/init/Config.java
```diff
@@ -109,6 +109,7 @@ public static class LaserArmConfig {
         public long maxEnergyInsertion = 512;
         public long energyPerTick = 128;
         public int blockBreakEnergyBase = 1024; // multiplied by block hardness
+        public float damageTickBase = 2;
```
Rearth, line 112: This also needs a translation key

Rearth (changes requested):
Hi, thanks again for the changes. This is already looking much better. However, it seems like the changes from my PR to your branch were lost, was this intentional? This now means that the laser arm doesnt track entities on the client again properly, and the rayCast checks arent working.

src/main/resources/assets/oritech/lavender/entries/oritech_guide/interaction/enderic_laser.md
jshipley: I updated the manual to fix the bolded header and mention the affect of speed addons.

src/main/java/rearth/oritech/block/entity/machines/interaction/LaserArmBlockEntity.java
jshipley: I reorganized this so that it will always call updateNetwork() if the network is marked as dirty as long as it's server side and the laser arm is assembled.

jshipley: I did notice that the "isActive" check doesn't seem to be working, though. The laser arm is getting charged even before it's assembled. I also checked the assembler, and it was also getting charged before it was assembled.

jshipley:
I'm not sure what happened to your changes. I was pretty sure that I had already merged them into my branch.

Merged (again?), and requested changes have been made.

I'm really looking forward to the possibilities of setting up a wither or shulker farm using a spawner controller, enderic laser, and in the case of the wither, reinforced deepslate.

Rearth (approved):

Rearth:
Sorry for the delay and all the back and forth. But I just loaded up your branch and am really happy with how it turned out. Huge thanks for all the work on this. I'll go ahead and merge it for now

PR #90, opened 2024-09-19
Adding compatibility recipes for Tech Reborn and Energized Power

jshipley:
Added some basic compatibility between Oritech and Tech Reborn/Energized Power.

Some highlights:
* Ore compatibility in pulverizers/grinders
* Alloying compatibility
* Added Oritech assembling recipe to Energized Power assembler (not the other direction, Oritech doesn't currently support using more than one item out of an ingredient stack)
* Added Oritech ores to output of Energized Power's Filtration Plant
* Added Fuel Generation recipes for Tech Reborn's advanced fuels.
* Added a couple of tags to recipes to allow for items to be usable in more crafting recipes. Example: any Energized Power wires are usable in place of Oritech's Fine Wire.
* Reorganized recipes into directory structure

What this does **NOT** do:
* Attempt to duplicate all of the recipes from other machines
* Allow machines from other mods to produce Fluxite/Enderic Compound

Overall, I think that these new recipes should make it more comfortable to use Oritech with the other major tech mods available on Fabric while still leaving the Oritech machines and the other mods' machines with their own specialties.

----

I also attempted to move the Alloy Forgery recipes into datagen. Trying to add Alloy Forgery to dependencies and import it for datagen completely broke datagen. Alloy Forgery was running its own datagen and producing all of its recipes, and none of the Oritech recipes were being produced. I ended up deciding to leave it as-is.


Rearth:
Damn, this must have been a ton of work. Huge thanks, this is looking great

PR #91, opened 2024-09-22
Fix NBT handling for dropped storage items

jshipley:
Avoid adding empty custom data to dropped items, and avoid null pointer crash if custom data is empty.


Rearth:
LGTM, thanks for the fix

PR #97, opened 2024-09-26
Expand chainsaw/drill/prometheum tools enchants/uses

jshipley:
* Removed tools from most of the tags that gets them enchants
* Added tools to relevant swords, axes, pickaxes, and shovels tags, which should get them all of the relevant enchants
* PlayerEntity will also treat chainsaw/prometheum axe as sword when attacking (hitting more than one mob at a time)
* Chainsaw/prometheum axe will break cobwebs and other sword efficient blocks like a sword

The chainsaw/prometheum axe were not made to handle hoe mineable blocks, and the drill/prometheum pickaxe were not made able to pave paths or do other shovel actions. As discussed, it seems fitting to not add that functionality to the power tools. Besides, it's cheap/easy to craft hoes/shovels for those finer tasks, and even stone tools are good/fast enough.


Rearth:
This is looking great, thanks. Not sure if I want to add too many mixins to this mod, but that a whole separate issue.

jshipley:
I get it. I was trying to avoid mixins, but there were some things that I couldn't find any cleaner way to do it. I have made an effort into keeping the mixins as non-invasive as possible so that they hopefully won't conflict with other mods or be difficult to update when Minecraft changes.

--
Jeff Shipley

Issue #92, opened 2024-09-22
Mythic Metals Compatibility


OakPlatinum:
### Describe the feature you'd like

Add compatibility with [Mythic Metals](https://modrinth.com/mod/mythicmetals).

Process ores and ingot from Mythic Metals with Oritech machines.

### Describe alternatives you've considered

Add new types of clumps, small clumps, gems, dusts, etc.

jshipley:
Adding recipes (fragment forge for splitting ores into raw metals, foundry for alloying, etc.) for Mythic Metals would be easy, and could probably be included in #90 (or another PR if that gets merged before this is ready). I'll spend a bit of time exploring Mythic Metals over the next day or two to see what would be most useful and add some compat recipes.

Full compatibility with dusts, clumps, etc. added would be a lot more effort and would involve adding a lot of art and conditional registration and is unlikely to happen.

jshipley:
I took a look, and there's really not much to add for Oritech. Other than splitting ores into raw metals with a fragment forge, I could also add alloying recipes for bronze, durasteel, star platinum, and steel.

The more advanced alloys that require 3-5 ingredients would have to be done with Alloy Forgery (which is a required dependency for Mythic Metals anyway).

PR #98, opened 2024-09-27
Adding some recipes for Mythic Metals compatibility

jshipley:
For #92

Added compatibility recipes for Mythic Metals and moved Alloy Forgery compatibility recipes into datagen.


OakPlatinum:
Great! Thank you!
Btw could you add "close #92" in the commit message so it automatically closes that issue?

Rearth:
This is looking great, thanks for all the worr that went into this. When merging i'll adjust the message to close the issue with this. One more thing before merging: Is there a way to only include those mobs in datagen, and maybe turn them on and off when actually running the dev environment? I don't want to have all  compat mods active when working on the mod, and having to manually enable them and reload gradle before each datagen also doesnt seem right?

jshipley:
I'll see if I can add the dependencies in a datagen run configuration.

jshipley:
That was easy. I tried adding a dependencies block inside of configureDataGeneration and the datagen is still able to run. It also makes it more obvious what the dependencies are for and prevents anything else from accidentally relying on them.

Rearth:
Thanks, this is looking cleaner already. I'll go ahead and merge it

PR #99, opened 2024-10-03
Several small fixes and improvements

jshipley:
Let me know if you would like these separated into several small PRs (or if you'd prefer that in the future). Any of these can be pulled out or improved if you'd rather not have the changes.

* New advancements
    * Level 10+ enchantment
    * Acquire gold gem
    * Steam engine
    * Ultimate core
    * Visit resource node
* UI changes
    * Enchantment Catalyst: shift-click an enchanted book puts the book in the book slot, and anything else in the other slot
    * Item Filter: shift-click an item adds it to the filter, unless it's already in the filter
    * Item Filter: setting default to whitelist
    * Atomic Forge: shift-click an item will fill the left (main) slot first, then the two right (secondary) slots. This was done by making the left slot 0 instead of slot 1, and updating recipes to match
* Spawner controller can now capture shulker when it attaches to the top of the controller (easiest to do by placing spawner controller underneath shulker attached to wall, then breaking wall)
* Comparator output for fluid/energy tank changed. Was emitting 0 until 1/15 full. Now emitting 0 for empty, 1 until 1/15, etc.
* Item pipe hitbox added because default pipe hitbox was larger than item pipe
* Fluid tank now emits light when containing lava
* bananas added to loot from jungle leaf decay (same probability as apples from oak)
* bananas and packed wheat are now compostable
* Added several compostable items to BIOMASS tag
* biomass item is now edible (like a poisonous potato)
* biopolymer can be made with solid biofuel
* flowers (and lapis) can be pulverized into dyes
* pulverizer can turn cobble -> gravel -> sand
* dirt can be crafted with biomass and sand
* dirt can be turned into clay in fluid centrifuge


Rearth:
Wow these are huge and great changes. Thanks for all the work

PR #103, opened 2024-10-04
Add Some Creative Blocks

OakPlatinum:
Creative blocks are useful in many aspects; they can be used for debugging, mapmaking, etc.
This PR added some creative blocks, including:

- [x] Creative Energy Storage
- [x] Creative Fluid Tank


OakPlatinum:
Hi, @Rearth. I have a problem. I try to refill or empty every tick, and it indeed works in filling/emptying buckets, filling the tank downside and fluid pipe transfering. However, GUI screen doesn't work; it just keeps displaying it's empty. I'm not sure how to fix it and don't want to mess up other codes. Could you help me?

Rearth:
Hi, thanks for all the work. I think the issue is related to the networking not being updated. Whenever you make any changes that should be synced to the client, make sure you set "netDirty" to true in the tank entity. Although I'm not sure if we really need the gui for the creative blocks, as well as the custom drops. I think the easiest way would be to just set the fluid type of the tank whenever a player right-clicks it with a bucket.

OakPlatinum:
图片
I tried adding markDirty before return, but gui still works wrongly.

Rearth:
Hm. Maybe the numbers are too large. I remember there being a conversion somewhere from fabrics units system to millibuckets. Maybe try it with a much lower capacity

OakPlatinum:
I've tried, and it doesn't display negative number now. Thank you!

Rearth:
Just took a look at this, and it's looking great. There's a few things in the new classes that aren't really needed, but I'll iron those out after merging. Huge thanks for all the work, this is something Oritech was really missing.

OakPlatinum:
Yeah, I knew some of the codes were unnecessary, but I haven't thought of changing the original small fluid tank entity class. It makes more sense now. Thanks for the correction!

PR #107, opened 2024-10-12
zh-CN Translation for Book

OakPlatinum:


OakPlatinum:
@Rearth Hi. I'm just wondering when v1.0.0 will be possibly released. There's still much additional work undone (eg. formatting, proofreading) for the translation of the book, but I'd like to have the Chinese translation included in the first stable version. If soon, I would possibly commit them first and fix in the later versions. Thanks!

Rearth:
Hm, I cant really give you an estimate here. I think Oritech itself is pretty stable right now, so feature/polish wise there's not much missing. I currently plan on getting started with the multiloader migration next (e.g. allowing oritech to be available on neoforge and fabric). For that I have to first move all logistic related logic to common storage lib, and then get started on the neoforge version. I have no idea how long these 2 tasks will take. With forgified fabric API it probably won't take too long, but with additional testing and stuff it's a few weeks at least. So as a very vague estimate i'd say 1-3 months, but it could be much earlier or later. I'm fine with merging the changes now and then doing improvements, or we wait a bit longer and then merge your polished changes. I think we also have the time for this option, but it's your choice

Rearth:
Closing this for now as the books have been removed in favor of the online wiki

PR #108, opened 2024-10-13
Fixed some recipes and images

OakPlatinum:
When I was translating the book, I found some recipes and images broken.

This PR aims to fix these issues, and it won't be set to draft so that you can merge it anytime and fix them yourself if you want.


Rearth:
Oh thanks, yeah we missed those when renaming stuff in datagen for compat reasons. Good catch

PR #110, opened 2024-10-16
Prometheum pickaxe additional blocks depend on mining direction

jshipley:
The direction the player is mining is now taken into account when choosing which additional blocks to mine. For example, if the player is mining the top of a block, the block in front of and behind (from the player's perspective) the mined block will be broken rather than always breaking the block above and below the mined block.

To make that work, the additional block breaking was moved to preMine (Block.onBreak) from postMine (Block.onBroken).

Also, the item enchantments now affect the block drops for the additional blocks.


jshipley:
It looks like this failed while trying to download java to set up the build. Is it possible to trigger it again, or do I need a new commit on this branch to trigger the build again?

Rearth:
The failing build seems to be a temporary networking issue, you can ignore that for now.
Just tested your changes locally, and they're working great. Hoge thanks for the improvement. One thing I'd still like to add is updating the outline of the targeted block to also show the other block that will be broken when in area mode. Although this is just me draming and doesn't actually have anything to do with this PR. Huge thanks again.

jshipley:
You're right, drawing the outline should be in a different part of the code and it would be really nice to have. I can take a look at adding it in another PR.

PR #112, opened 2024-10-19
Black Hole Tweaks

OakPlatinum:
This PR adds some configurations for the black hole block.


src/main/java/rearth/oritech/Oritech.java
```diff
@@ -19,6 +19,7 @@
 import rearth.oritech.client.init.ModScreens;
 import rearth.oritech.client.init.ParticleContent;
 import rearth.oritech.init.*;
+import rearth.oritech.init.OritechConfig;
```
OakPlatinum, line 21: I believed it was unnecessary, but my stubborn IDE insisted on this.

src/main/java/rearth/oritech/init/Config.java
```diff
@@ -60,6 +60,12 @@ public class Config {
     public int endPortalRequiredSpeed = 10000;
     public int netherPortalRequiredSpeed = 5000;
     public int blackHoleRequiredSpeed = 15000;
+
+    @SectionHeader("blackHole")
+    public boolean enableBlackHoleBlock = true;
```
Rearth: I don't think we actually need this part. The item is not sustainable in survival (e.g. on servers), and in creative I believe the players should just be able to do what they want. The item exists just for players to be able to build it in creative.

src/main/java/rearth/oritech/Oritech.java
Rearth: Yeah most IDEs don't seem happy with owo config

Rearth:
Hi, thanks for the PR. Making this configurable is a good idea, although I don't like the option to fully disable the black hole. It's not obtainable as an item in survival, you can only create the block directly. It's something that's supposed to be dangerous. Also it'll be needed for future things in the Oritech tech tree, including the block destruction part.

OakPlatinum:
It makes sense to me. Okay, I've removed that config so you should be able to merge it now.

Rearth:
Great, I'll go ahead and merge it then. Huge thanks for the PR again.

PR #133, opened 2024-11-01
stop work animation when tree cutter runs out of energy mid tree

jshipley:
I noticed that the tree cutter tick method always terminates early when there's not enough energy to continue cutting a tree, and thus keeps playing the working animation and sound even though it's not working.

Now the tick method will skip searching and cutting if there's not enough energy, but will still update the animation and run the dirty check.

Is it a problem for the same animation to be repeatedly triggered? It was already doing that if the cutter didn't find any blocks to cut.


Rearth:
Thanks, yes animations can be triggered over and over again. If it's the same animation as the one currently playing, it won't change anything.

PR #138, opened 2024-11-01
render target outlines for pickaxe area mode

jshipley:
Outlines should now be drawn around all of the blocks that will be broken when a prometheum pickaxe is in area mode.

This was done by adding a new outline renderer to Fabric's BLOCK_OUTLINE render event.


Rearth:
Cool stuff. I tried doing this before at some point but failed. Using the event is a good approach, huge thanks.

Rearth:
In the future we might want to make the pickaxe with in different direction too, and then have it harvest a 3x3 area. With the targets shown before this would now match the game

PR #139, opened 2024-11-01
add voxel shape for item filter

jshipley:
The item filter has a voxel shape now. Chests underneath an item filter can now be opened, and it is easier to target containers connected to item filters.


Rearth:
huge thanks for the PR. This is a great thing to have

PR #141, opened 2024-11-02
Small Code Improvements

Supyovalk:
Some IF-ELSE Invitations and constanting (Making two range variables per weedkiller thread is useless)


Crazyminer34:
It also causes permanent crashing on a world if you spam click them lol

Supyovalk:
> It also causes permanent crashing on a world if you spam click them lol

Wait, how? Is turning the variables into Static final overloads them? but aren't acesses-only?

Rearth:
Hi, thanks for the PR. The crash crazyminer got probably isn't related to this. However, the files have been moved already, so I'll close this for now. Performance wise the variable initialization order makes 0 difference, as the compiler is pretty good already. Moving this to config values would be a better idea here

PR #154, opened 2024-11-10
translation of the wiki into French

kikipunk:
translation of the wiki into French
I also added 2-3 images


Rearth:
Wow that's fast, I just published the wiki yesterday. Huge thanks for all the work that went into this, and also for adding the missing images.

PR #158, opened 2024-11-15
Update hand_drill.json and chainsaw.json

Crazyminer34:
Updated the hand drill and chainsaw to change the item type from generated to handheld


Rearth:
H, thanks for the PR. This is definitly a great idea. However the jsons you edited are in the generated folder, meaning they're built by datagen and manual changes there will not be kept. I just included your changes in the datagen code instead: 3e2b11c55e5a8de33bbf090fa8b4861edff22c1a
I hope it's okay with you that I directly included the changes in the datagen instead of waiting for you to update the PR.

Crazyminer34:
Lol, I never realized I wasn't editing the datagen, i had just assumed they were json files.

PR #168, opened 2024-11-20
Texture and lang fixes

Crazyminer34:
Fixed the word dubios across all files and file names to be called dubious instead (correct spelling)
Modified nickel and Platinum resource node textures to better match current texture design
Fixed english lang file having the redstone resource node as "Resource Node Redstone" instead of the opposite way around


Crazyminer34:
During testing, I figured out that there is something else with an incorrect name and it is causing the dubious container texture to be broken

Crazyminer34:
Ill put this PR on draft till I find a fix

Crazyminer34:
I think it was a misnamed file causing issues, testing now

Crazyminer34:
It was an incorrectly named file

Rearth:
H, thanks for the PR. Now in all honesty I don't really see the point of renaming the dubious container, as that'll break pending translations from crownid. The in-game displays are already all translated, and in the code I don't care about it. The rename would also change the item id, so it'd be a breaking change for existing worlds and advancements. I'd avoid that if there's no real benefit to it. I'd be happy to take a PR for the resource node changes, those new texture sure do look better / more consistent

Crazyminer34:
I'll wait for the other guy who was working on textures to send his because they do look slightly better

Issue #173, opened 2024-11-27
Using Oritech reduces Texture Quality (MipMap)


Plompi:
### Issue description

When having the Oritech Mod installed and playing the game, you can instantly notice the reduced Texture Quality of every Block in the distance. I have prepared 2 images to demonstrate this behaviour. The first one is without Oritech installed.
![Without](https://github.com/user-attachments/assets/4bc27809-8bcf-4207-a55e-4247807cedc9)
![Oritech](https://github.com/user-attachments/assets/697ea01c-7265-4525-9bf7-24dbc61dc64b)

### Steps to reproduce

1. Create a Minecraft Instance with the Oritech mod and its dependencies installed
2. Enter a Minecraft World and oberserve the poor Video Quality

### Minecraft version

1.21.1 (Latest)

### Platform

Fabric

### Loader version

0.16.7

### Fabric API version

0.110.0+1.21.1

### Oritech version

0.11.3

Plompi:
I found the texture thats causing these issues. Its the **particle_collector_block.png** texture in the **common/src/main/resources/assets/oritech/textures/item/** folder. Changing it to be a multiple of 4 fixes the issue

Rearth:
Fixed with #174 , thanks for the report and PR

PR #174, opened 2024-11-27
fixed the texture size of particle_collector_block.png

Plompi:
fixes issue #173


Rearth:
Great, huge thanks. Will include it in the next release

PR #175, opened 2024-11-28
Making the wall mounted and hanging celling lights have no collision

Crazyminer34:
I dislike that they have collision because it makes compact builds more painful


Rearth:
Hi, I'd prefer not to fully disable the collision of them, as I know people will complain about that then. However I tookyour PR as inspiration and made the ceiling light collision block slightly smaller, allowing players to walk through them. I think that's good enough for this issue now.

Issue #180, opened 2024-12-01
All dye recipes result in white dye


MuteTiefling:
### Issue description

All the pulverizer dye recipes are giving white dye instead of the intended color

### Steps to reproduce

1. check recipes in recipe viewre

### Minecraft version

1.21.1 (Latest)

### Platform

Neoforge

### Loader version

21.1.81

### Fabric API version

na

### Oritech version

0.12.1

### Other relevant versions

https://github.com/Rearth/Oritech/blob/f8062dc79fb1e8a13126fd48f40989d6f89e66ef/common/src/main/generated/data/oritech/recipe/pulverizer/dust/dyes/red.json#L15

tbouvencourt:
I’m currently working on solving this. [Message for other contributors]

Rearth:
Fixed with c0d098baec9f09ca400d110db7fd3217b2345683

PR #184, opened 2024-12-03
First contribution (and certainly not the last), aiming to fix: All dye recipes result in white dye #180.

tbouvencourt:
Change the JSON file as mentioned in the issue.

Then follow this idea and replace the white dye copy-paste with the correct color:

"Should be a pretty simple fix (copy-paste error). I can't see why they wouldn't accept a PR to fix this unless they already have it fixed in their dev environment."

Reference: [Link to the code](https://github.com/Rearth/Oritech/blob/046bea2ea92b01059b7349fdd50553e6c090ce27/fabricdatagen/src/main/java/rearth/oritech/fabricgen/datagen/RecipeGenerator.java#L109)

### DIFF
--- a/common/src/main/generated/data/oritech/recipe/pulverizer/dust/dyes/black.json
+++ b/common/src/main/generated/data/oritech/recipe/pulverizer/dust/dyes/black.json
@@ -12,7 +12,7 @@
   "results": [
     {
       "count": 1,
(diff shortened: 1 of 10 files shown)


Rearth:
H, thanks for the PR. You should be able to avoid having to manually edit the files by running the datagen steps, but this works too.

PR #193, opened 2024-12-07
Fix Invalid Pipe Updates

aderoian:
### Problem
After a connection between pipe and machine was disabled via Wrench, if that machine's blockstate was updated, the pipe would re-connect to that machine. This is __not__ intended behavior.

### Fix
The neighboring pipes to a machine are now stored in persistent data. Only when a new machine is being placed in a world will a pipe connect.

### Backwards Compatibility
This patch __is compatible__ with older versions. Once a pipe node is updated (toggling a connection will update a pipe node) it will trigger any neighboring machines to create a map of that machines neighboring pipes.


Rearth:
Cool stuff, huge thanks for the work here. Is there a reason to go with a `HashMap<BlockPos, Set<Direction>>` instead of a Set<Pair<BlockPos, Direction>>? I feel that might a bit easier to understand as it's one less container level less. Although you could argue that the pair fills a similar role. The Pair would be a pair of machine position and direction (e.g. a direction and target position for a pipe connection).
But I'm also happy with the current solution. I'd be happy to merge it like this aswell. Let me know how you want to proceed here

aderoian:
In my head I wanted to map a BlockPos (representing a machine block) to all its neighboring pipes, hence `Map<BlockPos, Set<Direction>>`. At the end of the day there is no major differences between the two and I don't mind changing the implementation.

Rearth:
Looks great, huge thanks for all the debugging and additional work here

PR #202, opened 2024-12-09
Fix #200 - Crash When Placing a Comparator Reading a Large Energy Storage

### LINKED ISSUE #200: Crash When Placing a Comparator Reading a Large Energy Storage
### Issue description

Game crashes when a comparator is placed reading a large energy storage.

### Steps to reproduce

1. Place and assemble a large energy storage
2. Place a comparator so it "reads" the large energy storage
3. Game crashes

### Minecraft version

1.21.1 (Latest)

### Platform

Fabric

### Loader version

0.16.9

### Fabric API version

0.105.0+1.21.1

### Oritech version

0.12.3

### Other relevant versions

_No response_

https://mclo.gs/prTC8V4

aderoian:
Fixes #200 

### Changes
- Added `ComparatorOutputProvider` to standardizing getting a comparator value from a block entity
- `RedstoneAddonBlockEntity.RedstoneControllable` overrides `ComparatorOutputProvider#getComparatorOutput` and always returns `0` since only the redstone addon should be used to read comparator output.

### Fix
Initial Cause: Invalid casting [in SmallStorageBlock](https://github.com/Rearth/Oritech/blob/c04d9bffc5c9f6732a78a07d969c7c458d4767ec/common/src/main/java/rearth/oritech/block/blocks/storage/SmallStorageBlock.java#L82)

We now cast to `ComparatorOutputProvider` since all block entities of a block with a comparator output should be its child. Since `RedstoneAddonBlockEntity.RedstoneControllable` overrides `ComparatorOutputProvider#getComparatorOutput` there is no longer an invalid cast.

### DIFF
--- a/common/src/main/java/rearth/oritech/block/blocks/accelerator/AcceleratorSensorBlock.java
+++ b/common/src/main/java/rearth/oritech/block/blocks/accelerator/AcceleratorSensorBlock.java
@@ -15,6 +15,7 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
 import org.jetbrains.annotations.Nullable;
 import rearth.oritech.block.entity.accelerator.AcceleratorSensorBlockEntity;
+import rearth.oritech.util.ComparatorOutputProvider;
 
 import java.util.List;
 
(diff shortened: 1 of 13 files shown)


Rearth:
Great PR, thanks. Adding an interface to standardize this across entities is something I should've done much earlier already

PR #209, opened 2024-12-14
Improved Drones

aderoian:
# Changes
- Added 2 addon clips to the drone port
- Drone ports will now transport fluid with a fluid addon
- Drone ports interact with the redstone control addon
- Updated drone wiki page to reflect changes

# Fixes
- Material can no longer be inserted into a drone port if it is receiving a payload

# Follow-Up
- Redstone control addon should have dynamic settings provided by the respective machine
- Internal: Abstraction of machine classes should be reworked, DronePortEntity includes a lot of redundant code because in cannot extend an abstract machine class


Rearth:
Just did some testing with this, huge thanks for all the work here. I noticed that the fluid tank of the port isn't persisted yet, but I can fix that after merging

PR #212, opened 2024-12-16
Fix Resource Node Generation

aderoian:
## Problem
- Resource ore nodes were not generating in specific biomes

### Cause
- The surface boulder generator checked specifically for a dirt tag or sand tag base and only air above causing many attempts to generate the feature to fail. One notable biome being any biome with snow

## Changes
- Resource ore nodes now generate in all biomes
- Resource ore nodes can now generate underwater
- Resource ore nodes generate during the `Underground Decoration` step (Primarily to allow surface modifications to happen around boulders, i.e. trees, snow, surface features)
- Decreased the chances of boulders spawning due to more successive generation

## Follow-up
- Placement chances may need to be altered


Rearth:
LGTM, huge thanks. 

Just did some testing with this in my dev environment. After flying around in spectator mode at max speed for a while I think the spawn rates are a bit too low in this change. I already lowered them before you started this PR, so they are already lower that the latest release. I'll do some fiddling around with the values here. I don't want to wander around forever looking for the right nodes, and when walking on the ground instead of flying in creative it's already a lot harder to find them.

PR #213, opened 2024-12-16
Fix More Invalid Connection Pipe Updates

aderoian:
This fixes connection pipes not updating their state when a neighbor machine is removed.


Rearth:
Ah good catch, thanks

PR #222, opened 2024-12-24
Tier Addons

Lumengrid:


Rearth:
Hi, thanks for the PR. However, while I do appreciate the effort, I sadly will not be merging this. Tiered addons don't really fit the vision I have for Oritech. Having huge machines with tons of addons is the desired result here. Oritech avoids to have any direct tiers in most cases.
In the future I'd highly recommend checking in before working on new features, especially ones that change the design / direction of oritech.

PR #223, opened 2024-12-25
Creative tab for Addons

Lumengrid:


Rearth:
Hi, thanks for the PR. Again, I appreciate the effort, but creating a separate tab just for addons feels a bit weird. I do prefer to have the blocks all in the same tab, as I don't really like mods that add a new creative tab for every small thing. If we separate addons, we'd also need to separate generators, particle accelerators, etc.
I'll probably go in a similar direction but with the item group sections of owo lib, where items are still in the same tab, but you have categories on the left to further filter them

PR #226, opened 2024-12-26
Add laser compatibility with ae2 certus quartz cluster

VasariRulez:
Add PlayerEntity to getLootDrops and getDroppedStacks when laser finishBlockBreaking to add compatibility with ae2 certus quartz cluster


Rearth:
Hi, is this really something that should be changed on Oritechs side? From what I'm understanding here, the ae2 certus quartz cluster only drop when a player is in the loot context builder? That sounds like a bug that should be fixed on their side to me, no?
But I'm also fine with merging this PR. If having players as optional in the loot builders is something used not just for this one thing, then we should definitly merge this.

VasariRulez:
Hi,

From what I've seen in the AE2 code, they specifically check for an Entity in the LootParams Builder. It looks intentional and not a bug on their end.

image

I can't say for sure how many other things would benefit from this PR since I only looked at this interaction with AE2.

In the end, it's up to you to decide whether to merge this PR or not.
Have a nice day!

Rearth:
Alright, thanks for digging into this issue on the AE2 side. I appreciate it.

PR #227, opened 2024-12-26
Modified item transfer logic in ItemPipeInterfaceEntity for improved handling of multiple stacks

VasariRulez:
Modified item transfer logic in ItemPipeInterfaceEntity for improved handling of multiple stacks.
it now gets all possible stacks that it can transfer and cycle on them to stop on the first valid transfer.
it can hit the tps with bigger chests, like the netherite double chest from sophisticated


Rearth:
Hi, the item pipes only moving the first found item is desired here. It's much more performant and also avoids a few issues. With the change proposed here, we'd probably run into potential performance issues with larger, full chests. I intend oritech pipes to just be a simple but performant solution for simple use cases. There's plenty of good pipes mods for more complex uses already. Moving only the first item is a drawback I'm willing to accept for this. It encourages players to properly design their filters / targets to avoid items blocking the queues here.

PR #231, opened 2024-12-29
fixed #230 (solar panel producing energy despite having blocks above)

### LINKED ISSUE #230: Solar Panels produce energy when blocks are above them
### Issue description

The solar panel still produces energy despite having a block on top, obstructing its panel. 
Screenshot (399)

### Steps to reproduce

_No response_

### Minecraft version

1.21.1 (Latest)

### Platform

Fabric

### Loader version

0.16.9

### Fabric API version

0.102.1+1.21.1

### Oritech version

0.13.0

### Other relevant versions

_No response_

_No response_

Plompi:
this implementation might not be the best, but it does what i'm expecting, when a block is above the solar panel

### DIFF
--- a/common/src/main/java/rearth/oritech/block/entity/generators/BigSolarPanelEntity.java
+++ b/common/src/main/java/rearth/oritech/block/entity/generators/BigSolarPanelEntity.java
@@ -70,14 +70,16 @@ public class BigSolarPanelEntity extends PassiveGeneratorBlockEntity implements 
     public int getProductionRate() {
         var baseRate = ((BigSolarPanelBlock) this.getCachedState().getBlock()).productionRate;
         var skyLightLevel = world.getLightLevel(LightType.SKY, this.getPos());
-        isFolded = world.isNight() && skyLightLevel < 12;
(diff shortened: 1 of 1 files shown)


Plompi:
currently this check only applies to the solar panel block itself (visually the solar panel has 6 panels, 5 of them dont react to blocks above them, even with this new implementation, so further development might be needed if this is also wanted)

Rearth:
Hi, thanks for the PR. I'm not sure how I want to proceed here, as this also leads to new issues. I personally don't really care if the solar panel really does a correct sun check. If players want to cheese it, I'm fine with that. 

The fact that solar panels rotate to face the sun makes this a lot harder, as sun may reach them early in the morning from the sides, so it should produce energy sometimes even if fully blocked. So the sky doesn't really have to be reachable for the solar panels to provide energy. The current implementation is a nice compromise where you need some sort of light, but other than that I'm fine with the solar panel working with blocks nearby. This makes it easier for the players to understand. Otherwise players would complain that the panel doesn't work because there is a block somewhere above it, even though most panel blocks are unobstructed from most directions.

PR #236, opened 2024-12-30
Add Framed and Duct Pipe Blocks

aderoian:
## Additions
- Added framed pipe blocks. 
These behave the same as normal pipe blocks but have a decorative frame around them
- Added duct pipe blocks. 
These are full-block pipes and only connect to other respective pipes. Cannot connect to machines.

## Internals
- Added `AbstractPipeBlock` to create another layer of abstraction.

## Additional Credits
NovaKatt, pipe duct textures


Rearth:
Hi, first of all huge thanks for the PR, this is already looking great. If you want I can go ahead and merge it already. There's a few things that should be addressed before a release, although I can also tackle those if you want (although the last one might take me a bit longer than you):

- Duct variants are missing recipes
- Lighting issues when normal pipes are next to framed variants (nothing big, I'd also be fine with releasing like this), see:

- Pipe network building issues with fluid pipes and the new duct variants (don't always work, sometimes they do, based on the order they're placed in?). Prints error "Yeah your pipe network likely is too long" for a 5 block network. See this setup not working (after replacing some blocks it works again, but only sometimes):

aderoian:
It should be good now. I don't know what happened with your local branch, but the images you provided seem outdated. The frame has an updated model and duct variants have crafting recipes.

Rearth:
Oh right. Just updated my local branch, and things are looking pretty good now. The last issue that I noticed is that the reverse recipes yield the wrong amounts:
image
But other than that it's looking great. I can also fix this issue after merging if you want.

aderoian:
I matched the reverse with the input: frames require 8 pipes and 1 plating, and ducts require 4 pipes, 4 stones, and 1 plating.

Rearth:
Hm, I mixed something up then. I'll merge it and do some more testing before the next release. Huge thanks

PR #247, opened 2025-01-14
DronePortEntity: TargetDesignator improvements

Lumengrid:
In order to implement the targetpos from a different tool you check the instanceof instead of the item itself


Rearth:
LGTM. I assume this is used for oritech things?

PR #262, opened 2025-02-02
Wiki Translation (zh-CN)

OakPlatinum:
Sorry for the delay, but I did say I’d get the translations up before the 1.0.0 release (just kidding ;) )


Rearth:
Wow this is amazing, huge thanks. I really appreciate the effort that went into translating the whole wiki.

PR #299, opened 2025-03-25
Don't destroy immature netherwart with crop filter. Fixes #298

### LINKED ISSUE #298: Crop Filter Addon Nether Wart support

It would be nice if the crop filter addon was able to treat the Nether Wart like crops similar to how it does wheat where it only harvests blocks that are 100% grown. 

We have a workaround where we have an observer looking at one of the last blocks and waiting until it is grown to send a Redstone signal to the destroyer block to start its run around until it gets to the one we have the observer on. I have attached our setup to show how our workaround works. (for the test we didn't have enough of Nether Wart to fill the farm but we had enough for a proof of concept).

Image

### Additional context

We are trying to autocraft some potions and we need a bunch of Nether Wart for that. We need a lot of Nether Wart for this so we were planning on building a large auto farm with Oritech. We have a workaround but it is not great. See the alternatives section.

jshipley:
The netherwart block isn't actually a crop block (probably because bone meal doesn't work on it).

This PR changes the immature crop test for the block destroyer and enderic laser when a crop filter addon is present so that immature netherwart will not be destroyed.

### DIFF
--- a/common/src/main/java/rearth/oritech/block/entity/storage/SmallTankEntity.java
+++ b/common/src/main/java/rearth/oritech/block/entity/storage/SmallTankEntity.java
@@ -36,6 +36,8 @@ import rearth.oritech.util.item.ItemApi;
 import rearth.oritech.util.item.containers.InOutInventoryStorage;
 
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


jshipley:
Cocoa blocks should also be accounted for. I can't think of any other minecraft croplike blocks that aren't subclasses of CropBlock.

jshipley:
It looks like this was closed accidentally when I reset my 1.21 branch (I forgot to "git checkout" after "git branch" when I made the new branch for this PR). I don't see any option to change the branch that this compares against.

Are you able to change this to pull the jshipley/netherwart_crop_filter branch, or should I just open a new PR?

jshipley:
I'm just going to open a new one from the correct branch...

PR #301, opened 2025-03-29
Add recycling tags/recipes

jshipley:
I added a bunch of new grinder/pulverizer recipes to recycle common materials (especially some building blocks and dungeon loot)

* Some things, like netherite tools/armor or copper blocks (except for "Copper Block" which can be unpacked) will lose a lot of material.
* Most metal tools/items will lose some material.
* Quartz/stone/sand should come out about even.
* Leather/leather equipment ends up as biomass. It made more sense coming out of a grinder than leather :)


jshipley:
I'm going to have to do enough rework of the recipes while I'm adding compatibility work that I'm going to include this with my compatibility changes instead of trying to merge it in later.

PR #320, opened 2025-04-14
Fix crash caused by accelerating particle

### LINKED ISSUE #318: Particle Accelerator will crash the game (including the server) if it kills the player while it is running
### Issue description

When a player is hit by an accelerating particle, it will immediately cause the game to crash, including the server.

### Steps to reproduce

_No response_

### Minecraft version

1.21.1 (Latest)

### Platform

Neoforge

### Loader version

21.1.145

### Fabric API version

None

### Oritech version

0.14.5

### Other relevant versions

ModPack: Ocean Block 2 1.7.0

_No response_

zeng-github01:
Fixes #318

### DIFF
--- a/common/src/main/java/rearth/oritech/block/entity/accelerator/AcceleratorControllerBlockEntity.java
+++ b/common/src/main/java/rearth/oritech/block/entity/accelerator/AcceleratorControllerBlockEntity.java
@@ -374,7 +374,7 @@ public class AcceleratorControllerBlockEntity extends BlockEntity implements Blo
         
         var maxApplicableDamage = mob.getHealth();
         var inflictedDamage = Math.min(remainingMomentum, maxApplicableDamage);
-        mob.damage(world.getDamageSources().magic(), remainingMomentum);
+        mob.damage(world.getDamageSources().magic(), inflictedDamage);
         var position = mob.getBoundingBox().getCenter();
         position = new Vec3d(position.x, particle.position.y, position.z);
         ParticleContent.BIG_HIT.spawn(world, position);


zeng-github01:
@jshipley What do you think of this PR? Is there a better solution? Is this an unexpected bug?

For example, accidentally passing the wrong function parameters

jshipley:
I haven't tested anything, but the change looks correct to me.

zeng-github01:
> I haven't tested anything, but the change looks correct to me.

What I am wondering is that after this change, the player's armor is not ignored. It is impossible to kill instantly under armor. Do I need to provide a new DamageSource that ignores armor? But I don't know how to implement it yet.

How was this originally intended to be implements?

The original code looks like it should be the same as it is now. It just passed the wrong parameters.

jshipley:
That's a good point.

The damage reduction is kind of weird. If you're doing 1000 damage and the player has 16 health. This will reduce the damage to 16, and then mob.damage() will further reduce it with shield blocking and armor. Then next time, if you're down to 8 and get hit with 1000 it will be reduced to 8, then further reduced by shield blocking and armor.

Do you have a stack trace for the crash? I've been reading through the LivingEntity code, and I'm not spotting why large amounts of damage would cause a crash.

zeng-github01:
> That's a good point.
> 
> The damage reduction is kind of weird. If you're doing 1000 damage and the player has 16 health. This will reduce the damage to 16, and then mob.damage() will further reduce it with shield blocking and armor. Then next time, if you're down to 8 and get hit with 1000 it will be reduced to 8, then further reduced by shield blocking and armor.
> 
> Do you have a stack trace for the crash? I've been reading through the LivingEntity code, and I'm not spotting why large amounts of damage would cause a crash.
> 
> 

I have used this patch privately. It's hard to withdraw. But I am willing to help solve this problem. But the problem is that I don't know how to build a high-damage particle accelerator.

If you can provide a development save, I will test it for you.

By the way, the particle accelerator that crashed was created by someone else in the multiplayer game.

I checked several logs. Not sure if it is this bug, it seems to trigger the watchdog. The crash occurs when the player clicks the "Resurrection" button

Rearth:
Hi, sorry for closing this PR, but without any logs that point to this actually causing any issues, I don't want to proceed here. The original code is working as intended, the variable names are just confusing. The remaining momentum is applied is damage, as the amount is being reduced by armor, so a high number is intended here. However the max damage that could be caused, so the remaining momentum is reduced by the var called "inflictedDamage"

zeng-github01:
> Hi, sorry for closing this PR, but without any logs that point to this actually causing any issues, I don't want to proceed here. The original code is working as intended, the variable names are just confusing. The remaining momentum is applied is damage, as the amount is being reduced by armor, so a high number is intended here. However the max damage that could be caused, so the remaining momentum is reduced by the var called "inflictedDamage"

Tested with the latest 1.21 branch, this problem no longer occurs. Never mind

PR #325, opened 2025-04-16
Machine frames will connect to any tag-configured frame support block.

### LINKED ISSUE #207: Make Machine Frames connect to the Industrial Support Beam

image
Would be nice if the Machine Frame would make a connection down to the Support Beam

_No response_

### Additional context

_No response_

jshipley:
Fixes #207

The only block configured to be a frame support block is the metal beam. Metal beams blockstates were also changed to treat machine frames above/below them as beams as long as the beam is in the frame support tag.

Adding machine frames to the frame support tag will also make them connect to each other vertically.

### DIFF
--- a/common/src/main/generated/data/oritech/tags/block/frame_support.json
+++ b/common/src/main/generated/data/oritech/tags/block/frame_support.json
@@ -1,0 +1,6 @@
+{
+    "values": [
+      "oritech:metal_beam_block",
+      "oritech:machine_frame_block"
+    ]
+  }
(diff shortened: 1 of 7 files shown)


Rearth:
Cool stuff. Just took a first look at the code, and I do like the idea of adding a tag for the beam connection supports. Could we further simplify this code by also just marking the frames themselves as support blocks, so we don't need to check for both the tag and the support blocks? Or do we want the support blocks to only work on the bottom side, and thus not allow this?

jshipley:
Are you talking about the metal beam class? 

The machine frames work great as supports and connect to each other vertically if they're in the tag. It's a little busy looking if two parallel frames are stacked, but can look great if they form a cube.

The more complicated bit is for the block states of the metal beams.  Basically, if the beam is a frame support, then the blockstate for the beam should act like machine frames are beams when it's deciding whether to display the brackets on top and bottom of the beam. 

I can review that and see if I can simplify it more.

jshipley:
It could be simplified if the beam was always a machine frame support block regardless of the tag. Other than that, I'm not sure how to get simpler than "beamAbove = (block above me is a beam) or (I'm a frame support and the block above me is a frame)"

PR #330, opened 2025-04-18
Add Laser minimum transfer rate

### LINKED ISSUE #327: Enderic Laser no energy transfer when Maximum Usage gt target Max Input
### Issue description

If `storageCandidate instanceof DynamicEnergyStorage`, it works correctly.
For other energy storage, it not ignore input limit and simulate to check `inserted == transferCapacity` before transfer.
When Maximum Usage > target Max Input, the simulate return inserted = target Max Input.
At this point, the two values are not equal, so no energy be transferred.

### Steps to reproduce

1.Target a Pump (Max Input = 1000) will energy transfer work normally.
2.Add some speed addons to increase Enderic Laser Maximum Usage over 1000.
3.Nothing energy transfer happen.
4.Remove speed addons to reduce usage less than 1000.
5.The Laser work again.

### Minecraft version

1.21.1 (Latest)

### Platform

Neoforge

### Loader version

21.1.127

### Fabric API version

0.107.0+2.0.24+1.21.1

### Oritech version

0.14.5

### Other relevant versions

_No response_

_No response_

jshipley:
Fix #327. Let the laser transmit any amount of energy between a configurable minimum and the machine's available storage.

When the laser has addons that make it capable of transferring more energy per tick than a machine can receive, it was failing to transfer anything.

A new laser arm config property was added that sets a minimum threshold for energy transfer.

The new default is 256, which is 1/4 of the laser's default max insertion rate and 1/2 of the basic tools' max insertion rate.

### DIFF
--- a/common/src/main/java/rearth/oritech/block/behavior/LaserArmBlockBehavior.java
+++ b/common/src/main/java/rearth/oritech/block/behavior/LaserArmBlockBehavior.java
@@ -82,7 +82,7 @@ public class LaserArmBlockBehavior {
                 
                 if (storageCandidate instanceof DynamicEnergyStorage dynamicStorage) {
                     var inserted = dynamicStorage.insertIgnoringLimit(transferCapacity, true);
-                    if (inserted == transferCapacity) {
+                    if (inserted > 0 && inserted <= transferCapacity) {
                         dynamicStorage.insertIgnoringLimit(transferCapacity, false);
                         dynamicStorage.update();
                         return true;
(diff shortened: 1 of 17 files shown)


Tctiancan:
It looks like these changes will cause normal laser (default transfer amount 128 is less than minEnergyExtraction 256) no longer transfer energy.
And transfer less than minEnergyExtraction energy to the player will no longer consume laser energy.

jshipley:
Yeah, I did the player bit wrong.

With the minimum being so low, I think I'm going to revert this and just not have a minimum. If you want to have a superspeed laser that's spending 10000 energy to give a pump 32 energy then more power to you :)

Rearth:
Hi, thanks for this PR. This is looking good in the current setup. Having a minimum transfer amount would not work with everything, as things like the atomic forge need a perfectly full energy storage to proceed. Wasting some RF is the way to go here.

PR #331, opened 2025-04-18
Allow tags in fluid ingredients

### LINKED ISSUE #311: Allow using fluid tags in recipes

Allow using fluid tags instead of fluids for fluid inputs in recipes

Adding separate recipes for each entry in a tag

### Additional context

_No response_

jshipley:
* Added a new FluidIngredient class that will accept either a fluid id or tag, plus a fluid amount.
* Added fuels to tags, and tags to recipes
* Moved Datagenerator Class to common, then called it from fabricdatagen and neoforgedatagen - now all recipes on Fabric will use Fabric fluid amounts and all recipes on Neoforge will use Neoforge fluid amounts. No extra math necessary, and people writing recipes can use the native fluid amount in their recipes as well.
* Made sure that recipe viewers were working with Fluid tags

The main reason I wanted this is to make it easier for modpack developers to add fuels to the fuel generators. New recipes are still an option (especially if you want a fuel that lasts shorter or longer than the oritech fuels), but adding fuels to the basic recipes can now be as easy as tagging the fuels.

Fixes #311 

For the recipe viewers, the plugins do different things but should be able to convey enough information. JEI changes the displayed fluid every couple of seconds, EMI picks one to render and then has a tooltip with a list (including small rendered square) of fluids in the tag, and for REI I added a displayed the tag in the tooltip and then added a (truncated) list of fluids to the next line of the tooltip. 

As usual, any changes to datagen end up with huge changelists. There were 20 or so java files touched, with most of them being small changes. It still ends up with several hundreds of changed files, though.

With the new endecs for fluid input (and output, to keep it consistent), the recipe format has slightly changed. The fluid sections should look a bit like this:

```
"fluidInput": {
    "fluid": "minecraft:water",
    "amount": 250
},
"fluidOutput": {
    "fluid": "minecraft:empty",
    "amount": 0
}
```

fluid can be an identifier like "minecraft:water" or a tag like "#c:oil". If amount is omitted, it should default to 1 bucket (as long as the fluid is not empty). "fluidInput" and "fluidOutput" should also be optional and default to empty. That won't affect datagen recipes which have all of the fields even if they're empty, but anybody adding their own recipes won't have to include all of the empty sections.


jshipley:
Ouch. 1300 files changed. A big part of that is the format of every recipe changing, and part of that is every recipe being duplicated for neoforge. Github really needs to let you hide files by subdirectory or extension for the reviews.

Outside of generated files, though, it's 22 java files changed and a couple of the build.gradle and gradle.properties.

Rearth:
Very cool, huge thanks for the work that went into this, and figuring out the codes. I really like the great recipe viewer integration. Amazing work.

PR #333, opened 2025-04-18
Fix #324. Copy inventory before every complexMatch attempt

### LINKED ISSUE #324: Energite Ingot has wrong recipe logical on latest commit
### Issue description

When making this item, more than 3 raw materials are required to make one, and only one raw material is actually consumed.

Revert da20a581a5d1e2218e2d8967f7b0110102123ff7 Can fix this

### Steps to reproduce

_No response_

### Minecraft version

1.21.1 (Latest)

### Platform

Neoforge

### Loader version

21.1.128

### Fabric API version

IDK

### Oritech version

Latest Github Commit

### Other relevant versions

_No response_

_No response_

jshipley:
Recipe matching removes ingredients from the copy of the crafting inventory with every attempted recipe match, causing recipe matching to fail because of missing ingredients.

Fixed by copying the input list before every match attempt.

Fixes #324

### DIFF
--- a/common/src/main/java/rearth/oritech/init/recipes/OritechRecipe.java
+++ b/common/src/main/java/rearth/oritech/init/recipes/OritechRecipe.java
@@ -79,12 +79,15 @@ public class OritechRecipe implements Recipe<RecipeInput> {
         
         // Input does not need to be in the correct slots / split into different slots.
         // We just check if we can remove all ingredients from the inventory, and fail is any input is not able to be removed.
+
+        // make a fresh copy for every search, as every recipe this attempts to match mutates the inventory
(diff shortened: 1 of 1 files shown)


zeng-github01:
You need to add Fixes #324 in the description of the PR, otherwise it will not be automatically closed after the merge.

Rearth:
Hi, this is something I'm still actively working on. The recipes were already being copied before matching (see https://github.com/Rearth/Oritech/commit/da20a581a5d1e2218e2d8967f7b0110102123ff7#diff-6195f2e538d70f4fb424156cfc7fd6c14470f8eefdc95839a7d70ccd25ed9665L85-L280), but some issues with there were still in the code, as not all calling functions ensured the inventory is copied. I already implemented something similar to this PR now, still thanks for the work that went into this.

PR #334, opened 2025-04-18
Netherwart crop filter

### LINKED ISSUE #298: Crop Filter Addon Nether Wart support

jshipley:
The netherwart block and cocoa blocks aren't actually a crop block, but should still be treated by the crop filter as if they are.

This PR changes the immature crop test for the block destroyer and enderic laser when a crop filter addon is present so that immature netherwart or cocoa will not be destroyed.

Fixes #298 

(This is a duplicate of the accidentally closed #299, but opened from a feature branch instead of the main 1.21 branch)

### DIFF
--- a/common/src/main/java/rearth/oritech/block/behavior/LaserArmBlockBehavior.java
+++ b/common/src/main/java/rearth/oritech/block/behavior/LaserArmBlockBehavior.java
@@ -4,12 +4,14 @@ import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.block.Blocks;
 import net.minecraft.block.CropBlock;
+import net.minecraft.block.NetherWartBlock;
 import net.minecraft.block.entity.BlockEntity;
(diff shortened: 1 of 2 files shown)


Rearth:
Hm, I kinda expected those to be implemented as crops, but oh well. Thanks for fixing them.

PR #338, opened 2025-04-22
Allow PlayerPortal to cross-dimensional teleportation

### LINKED ISSUE #335: Multiplayer  - PlayerPortal issue for the host
### Issue description

Hi there!

My wife and I playing together via LAN - she hosts the game, and I join. I’ve built a base with the Augmentation Workbench and all the addons, and we both share research and use the upgrades. Everything was going great, except for one issue we ran into with the Player Portal enchant.

For me, it worked perfectly—no problems at all. But for my wife, things didn’t go so smoothly. She could open the portal, but when she walked into it, I would disappear on her screen without actually being teleported. It was super frustrating!

We tried rebuilding the workbench and reinstalling the addons, but nothing seemed to fix it.

Since we really enjoy Oritech, I decided to dive in and learn a bit of Java to see what might be going on. After adding some logging to `PortalEntity.onPlayerCollision`, I noticed the event was being fired twice—likely once for the client and once for the server. On her PC, it looked like the client-side event happened first, which seemed to cause the portal to be removed before teleportation could happen.

Adding a check for `this.getWorld().isClient` helped resolve that issue. However, I ran into another problem: the portal doesn’t seem to handle multiple dimensions properly. Trying to teleport in the Nether caused some... unfortunate experiences 😅

To temporarily fix this for our playthrough, I made it teleport to the spawn location instead with  `serverPlayer.getSpawnPointPosition()` and `serverPlayer.getSpawnPointDimension()`  (which works with Nether Anchors too). It’s definitely not the intended logic, but it’s working for now while we wait for an official fix.

Thanks for the awesome mod—we're really enjoying it and looking forward to future updates!

### Steps to reproduce

1. Player 1 (host) visits base of Player 2 and installs PlayerPortal augment
2. Player 2 could spawn portal and use it
3. Player 1 could spawn portal but teleportation not working

### Minecraft version

1.21.1 (Latest)

### Platform

Fabric

### Loader version

0.16.12

### Fabric API version

0.115.4

### Oritech version

0.14.5 and 0.14.6

### Other relevant versions

_No response_

_No response_

fordarian:
Fix player portal augmentation in non-overworld dimensions (Fixes #335)

This PR (based on #337) fixes portal augmentations to work properly across all dimensions. Previously, only the coordinates of the augmentation center were stored without dimension information, causing issues when traveling between dimensions.

### DIFF
--- a/common/src/main/java/rearth/oritech/block/entity/augmenter/PlayerAugments.java
+++ b/common/src/main/java/rearth/oritech/block/entity/augmenter/PlayerAugments.java
@@ -26,6 +26,9 @@ import net.minecraft.sound.SoundEvents;
 import net.minecraft.util.Identifier;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Box;
+import net.minecraft.util.math.GlobalPos;
+import net.minecraft.world.World;
+
 import org.joml.Vector2i;
 import rearth.oritech.Oritech;
 import rearth.oritech.client.other.OreFinderRenderer;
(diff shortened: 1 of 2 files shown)


Rearth:
Cool stuff, thanks for the PR. This is looking great.

PR #339, opened 2025-04-22
Fix Addon tooltip

Tctiancan:
Due to floating-point precision issues, some addon display inaccurate values in tooptip.

image
image
https://github.com/Rearth/Oritech/blob/464265cea4c08b761bd772e9a7766e9f0eaafce6/common/src/main/java/rearth/oritech/init/BlockContent.java#L227
https://github.com/Rearth/Oritech/blob/464265cea4c08b761bd772e9a7766e9f0eaafce6/common/src/main/java/rearth/oritech/init/BlockContent.java#L232


Rearth:
Thanks for the PR, this has also been bothering me for a while, but never enough to actually look into the issue.

PR #341, opened 2025-04-22
Make Auto Feeder directly call the item eating function

### LINKED ISSUE #322: Autofeed ignores item durability or containers
### Issue description

Tested with beetroot soup and coffee cup from Actually Additions.

Eating beetroot soup should leave behind an empty bowl, but the autofeeder consumes the soup and the bowl.

Similarly, the coffee cup should have 3 uses and leave behind an empty cup. It gets consumed (cup and all) in on use, and doesn't apply any of the potion effects.

The autofeed should work more like actually eating the items rather then just getting the nutrition value then deleting the item.

### Steps to reproduce

_No response_

### Minecraft version

1.21.1 (Latest)

### Platform

Fabric

### Loader version

0.16.10

### Fabric API version

0.115.4+1.21.1

### Oritech version

0.14.5

### Other relevant versions

_No response_

_No response_

zeng-github01:
Fixes #322 

The new implementation directly calls the function of long pressing the right button to simulate the eating implementation as much as possible. At the same time, it restores the satiety with the least satiety waste as possible.

### DIFF
--- a/common/src/main/java/rearth/oritech/Oritech.java
+++ b/common/src/main/java/rearth/oritech/Oritech.java
@@ -107,7 +107,7 @@ public final class Oritech {
         res.put(RegistryKeys.ITEM_GROUP.getValue(), () -> ArchitecturyRegistryContainer.register(ItemGroups.class, MOD_ID, false));
         res.put(RegistryKeys.RECIPE_SERIALIZER.getValue(), ArchitecturyRecipeRegistryContainer::finishSerializerRegister);
         res.put(RegistryKeys.LOOT_FUNCTION_TYPE.getValue(), FluidContent::registerItemsToGroups);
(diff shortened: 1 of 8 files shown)


zeng-github01:
@jshipley Please review it. The item selection algorithm is from Copilot, so there may be errors.

Rearth:
This is looking mostly good. Can you update / rebase your branch with the latest changes? Also, did you verify that the algorithm works as intended ingame?

zeng-github01:
@Rearth The rebase has been performed, but it seems to have mixed my commits with another commit.

Never mind. Let's leave it at that.

As for the algorithm problem. I don't have time to test it now. I will test it when I have time.

By the way, I don't remember when I closed this PR...

It seems that it was automatically closed by rebase.

zeng-github01:
@Rearth I found that the content of this PR has been merged by you. That's it. I'll test it when I have time.

zeng-github01:
@Rearth The algorithm is working properly, there is nothing wrong with it. Since it is impossible to precisely control the player's hunger, we can only confirm that it does choose different foods.

The effects of Mystery Stew can also be granted normally.

PR #359, opened 2025-05-02
Adding biome modifiers to neoforge specific data files

jshipley:
Oritech features were not being generated on Neoforge (see https://github.com/architectury/architectury-api/issues/480 for other people having the same issue).

I added some neoforge biome modifier json files (https://docs.neoforged.net/docs/worldgen/biomemodifier/) that should do the same thing. I haven't tested all of the features, but I did verify that I was able to create a new world and find ore boulders.


Rearth:
Huh, I was pretty sure I tested the world gen on both neo and fabric when adding this, but yeah I was able to reproduce the issue. Thanks for the PR, this looks great, I'll also add a comment to the existing modifiers section so we remember to add the neoforge specific file when changing/adding things here in the future.

PR #373, opened 2025-05-04
Save Particle Accelerator (Controller Block) Inventory

Tctiancan:
image
-10, 79, 61 has the following block data: {x: -10, y: 79, z: 61, id: "oritech:accelerator_controller_block_entity"}
No inventory (Items) data, reopen game, items had vanished.
image


Rearth:
Good catch, thanks for the PR.

PR #395, opened 2025-05-10
Fix Neoforge datagen so all recipes go to oritech namespace

jshipley:
This is separated into two commits, which will hopefully make it easier to review.

Java file changes only:
https://github.com/jshipley/Oritech/commit/be8e1ff4eee961dd0235b90748118498235e2792

Generated JSON file changes only:
https://github.com/jshipley/Oritech/commit/64890b3c2ee476e35869677279fcf1581a65c327

## What changed

* Added Oritech.MOD_ID to the PackOutput (DataOutput), but it didn't seem to change much
* Added Oritech.id() to the recipe output wherever possible
* Copied the static offerSmelting/offerBlasting functions to the OritechRecipeGenerator class, made them use Oritech.id(), and used those instead of the RecipeProvider versions

There were some changes on the fabric side as well, but all of them I could see were in the advancements.

As an example, oritech/advancement/recipes/misc/crafting/treefeller.json had minecraft:crafting/treefeller as a condition/reward but now has oritech:crafting/treefeller as a condition/reward


Rearth:
Hi, thanks for this PR. This is looking mostly good. Having the changes separate like this is really helpful.

Rearth:
Did you test all those changes ingame?

neoforgedatagen/src/main/java/rearth/oritech/neoforgegen/OritechDataGenerator.java
```diff
@@ -24,7 +24,7 @@ public class OritechDataGenerator {
     @SubscribeEvent
     public static void gatherData(GatherDataEvent event) {
         DataGenerator generator = event.getGenerator();
-        PackOutput packOutput = generator.getPackOutput();
+        PackOutput packOutput = generator.getPackOutput(Oritech.MOD_ID);
```
Rearth: Are you sure this change is working as intended? IIRC I tried and it changed the folder structure to not be correct anymore

jshipley: You're right. I missed that the neoforge files were going to "oritech/data/oritech" with that.

The latest changelist (https://github.com/jshipley/Oritech/commit/26f9528dbdbc1a553e46ea167b2d023126ee1ab2) is just reverting this one line and rerunning the datagen.

I did play test after reverting this line, and I am seeing the recipes in neoforge.

Rearth:
Cool, thanks for the changes.

PR #403, opened 2025-05-11
Fix nugget output, disable compat ore recipes when jaopca is loaded

### LINKED ISSUE #381: [1.21.1] [Neoforge] Zing Centrifuge recipe giving wrong output
### Issue description

The fluid separator recipe for Zinc gives 2 zinc nuggets instead of 2 Zinc gems. 
Image

The separator should give 2 Gems when you use a Clump/Crushed ore and water 
Image

Having an Ore clump/crushed into a nugget feels like a waste, since they can smelt into a whole ingot. 

### Steps to reproduce

_No response_

### Minecraft version

1.21.1 (Latest)

### Platform

Neoforge

### Loader version

21.1.168

### Oritech version

0.15.2

### Other relevant versions

_No response_

_No response_

jshipley:
* Fixed nugget output for centrifuge recipes (it respects centrifugeAmount now)
* Added a not(modLoaded("jaopca")) condition to NeoForge compat ore processing recipes

Fixes #381

### DIFF
--- a/common/src/main/java/rearth/oritech/api/recipe/util/MetalProcessingChainBuilder.java
+++ b/common/src/main/java/rearth/oritech/api/recipe/util/MetalProcessingChainBuilder.java
@@ -275,14 +275,14 @@ public class MetalProcessingChainBuilder {
         if (clumpItem != null) {
             CentrifugeRecipeBuilder.build()
                 .input(clumpIngredient)
-                .result(firstNonNull(centrifugeResult, gemItem))
+                .result(firstNonNull(centrifugeResult, gemItem), centrifugeResult != null ? centrifugeAmount : 1)
                 .result(Optional.fromNullable(dustByproduct), byproductAmount)
                 .timeMultiplier(timeMultiplier)
                 .export(exporter, resourcePath + "clump/" + metalName);
             CentrifugeFluidRecipeBuilder.build()
                 .input(clumpIngredient)
                 .fluidInput(Fluids.WATER)
-                .result(firstNonNull(centrifugeResult, gemItem), 2)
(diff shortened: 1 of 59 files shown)


Rearth:
Looks mostly good. Do we have the JAOPCA results as compat output when it is installed or are there just no recipes now when it is installed?

jshipley:
The ore processing chain compat recipes are completely removed when JAOPCA is loaded, because JAOPCA duplicates it all and even adds new items like gems. It seemed unnecessary to keep the redundant recipes.

When I was looking around the recipes in game with Oritech, JAOPCA, Immersive Engineering, Create, and Mekanism installed, it looked like all of the ore processing recipes were provided by JAOPCA including grinding and centrifuging recipes.

Rearth:
Ah okay, good to know. Thanks for the work here then, this is looking great.

Staegrin:
This needs a slight tweak sadly. While JAOPCA does handle raw ore processing in the Fragment Forge and the Refinery.  It does not add ore block processing in the oritech machines or raw ores in the pulverizer. 
These are the only ways to process raw zinc with JAOPCA installed.


Only ways to process raw silver.




While without JAOPCA installed







oritech's own compat does provide ore block recipes when not turned off in the presence of JAOPCA.








In conclusion too many compat recipes are turned off in the presence of JAOPCA which in turn are not replaced by JAOPCA recipes leaving no way to process ore blocks and reducing ways to process raw ores.

Currently fixing this by removing 

```
    {
      "type": "neoforge:not",
      "value": {
        "type": "neoforge:mod_loaded",
        "modid": "jaopca"
      }
    }
```
from the affected recipes thereby reactivating the recipes.


Versions
minecraft 1.21.1
neoforge 21.1.219
oritech 1.1.0
JAOPCA 5.0.12.20

PR #410, opened 2025-05-16
Add iron helmet to recyclable/iron_dust.json

maximeurbn:


Rearth:
Hi, thanks for this PR. Sadly changing this directly in the json in the generated folder will cause this change to be overridden with the next datageneration run, as this is an automatically generated file. I just pushed a commit that fixes this directly, so I'll close this PR for now. Still thanks for the  PR. (see commit f098e391f27d0e856f1c4ba2b31e324f8fdb435a)

PR #423, opened 2025-05-25
Asset update

Crazyminer34:
Fixed Error in English Lang File
Modified Platinum and Uranium resource node textures to match their deepslate counterparts


Crazyminer34:
I was gonna upload and send this like 2 months ago but I forgot

Rearth:
Hi, those look great, thanks for that.

PR #431, opened 2025-05-31
Fixes energy api extract on fabric

### LINKED ISSUE #430: Incompat with Sophisticated backpacks [fabric port]
### Issue description

Something about an issue with energy api
See [this](https://github.com/Salandora/SophisticatedBackpacks/issues/136#issuecomment-2908654832)

### Steps to reproduce

_No response_

### Minecraft version

1.21.1 (Latest)

### Platform

Fabric

### Loader version

0.16.4

### Oritech version

15.7

### Other relevant versions

_No response_

_No response_

zeng-github01:
# Description

Fixed the problem that the Energy API implementation under Fabric did not update the insert and extract functions at the same time, which may lead to possible energy duplication.

Fixes #430 

# How Has This Been Tested?

This PR does not involve the common part, but only fixes the energy api of Fabric, so it is only tested on Fabric.

I have tested the scenario of this issue, used Portable Energy Storage and the battery plug-in to try to charge and discharge the backpack and Portable Energy Storage. No obvious abnormalities were found.

- [ ] Feature has been tested ingame on both neoforge and fabric.

# Checklist:

### DIFF
--- a/fabric/src/main/java/rearth/oritech/fabric/FabricEnergyApiImpl.java
+++ b/fabric/src/main/java/rearth/oritech/fabric/FabricEnergyApiImpl.java
@@ -187,7 +187,16 @@ public class FabricEnergyApiImpl implements BlockEnergyApi, ItemEnergyApi {
                     container.update();
                 }
             });
-            return container.extract(maxAmount, false);
+
+            long extracted = container.extract(maxAmount, false);
+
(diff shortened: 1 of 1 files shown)


Rearth:
Thanks for investigating and fixing this. The PR looks good.

Issue #422, opened 2025-05-24
[Suggestion] Allow Drag items from JEI/EMI to set items in the Item Filter


ffuentesm:
### Describe the feature you'd like

Just like create, it would be nice if we could set the item filters by using the items from the item list.
![Image](https://github.com/user-attachments/assets/d0fd59ff-a5f2-4968-814e-b030005100aa)

Rearth:
Fixed by  #483

PR #483, opened 2025-07-16
Item Filter Ghost Handlers

JustAHuman-xD:
# Description
Adds the relevant handlers for JEI, REI, and EMI allowing dragging items into the filter screen.
Added `ItemFilterScreen#FILTER_SIZE` to use instead of using 12 all over.
Added `ItemFilterScreen#acceptItemStack` and moved most logic from the click handler to it for common use.
Made the slot containers 18x18 instead of 19x18 and compensated by adding 1 to the margin instead.
Fixed the filter slots background only being 17 pixels instead of 18.

Moved `JeiExclusionZoneHandler` to a dedicated class file to match other recipe plugin standards, the ghost handler is in its own file as well.

Implements #422 

# How Has This Been Tested?
- [x] Tested Clientside Fabric
- [ ] Tested Clientside NeoForge
- [ ] Tested Dedicated Server Fabric
- [ ] Tested Dedicated Server NeoForge

# Checklist:
- [x] N/A ~~I have commented my code, particularly in hard-to-understand areas~~

### DIFF
--- a/common/src/main/java/rearth/oritech/client/ui/ItemFilterScreen.java
+++ b/common/src/main/java/rearth/oritech/client/ui/ItemFilterScreen.java
@@ -22,11 +22,13 @@ import java.util.function.Predicate;
 import static rearth.oritech.client.ui.BasicMachineScreen.ITEM_SLOT;
 
 public class ItemFilterScreen extends BaseOwoHandledScreen<FlowLayout, ItemFilterScreenHandler> {
-    
+
+    public static final int FILTER_SIZE = 12;
+
     private ButtonComponent whiteListButton;
     private ButtonComponent nbtButton;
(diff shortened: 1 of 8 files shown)


JustAHuman-xD:
I still need to do testing on neo forge & dedicated servers (but it shouldn't really matter) or if someone else wants to do it since I'm not sure when I'll have time!

Rearth:
Oh very cool, thanks for all the work here. I'll also check and test the changes locally in a bit.

Rearth:
Hi, just did some testing on neoforge aswell, and it's working pretty well. This feature is something I also wanted to add, but had no idea how. Huge thanks for the improvement, I'll include it in the next update/

PR #501, opened 2025-07-30
Migrate to Mojang mappings

rlnt:
# Description

This pull request migrates the project from yarn to Mojang mappings.

Mojang mappings are the official mappings provided by Mojang and are updated with each version.
They are widely used and are the standard for development with NeoForge.

The reason for this change is increased compatibility with other projects. A detailed explanation can be found here:
https://github.com/IrisShaders/Iris/issues/679

- reworked Gradle build scripts for better structure and extracted common logic
- remapped all modules to mojmaps
- switched to mojmaps with layered parchment
- adjusted access widener
- adjusted mixins
- fixed some issues manually that were not targeted by automatic remapping
- moved datagen from dedicated subprojects to their respective loaders
  - ensured they can still use dedicated dependencies
  - ensured the main source set doesn't load the datagen source set
- switched from volatile `@ExpectPlatform` system to service loader
- moved from Architectury Loom to MDG (Common and Neoforge) and Fabric Loom (Fabric)
- data generation has been split up between Fabric and Common
- add Maven publishing

# How Has This Been Tested?

~~The code doesn't compile yet because of a remapping conflict, which is caused by Architectury Loom. That's why it's currently listed as draft. A tracking issue has been opened here: https://github.com/architectury/architectury-loom/issues/292~~
~~The Architectury Loom issue is still a thing, but I was able to work around it.~~

~~Furthermore, since a lot of Gradle build script adjustments have been done, the data generation source sets are currently not activated to focus on the main issues of the code base.~~

~~The last issue remaining is that NeoForge can't find a class in owo-lib, which I have no explanation for yet. I need to investigate a bit more. Everything else is working.~~

~~When the outstanding issues have been resolved and an artifact can be built, I will test the new codebase with popular add-ons via Maven local check if they can still compile against the new mappings. After that, I am planning to add official Maven publishing as well, so it's easier for other developers to depend on the mod and get access to proper sources.~~

The data generation has been run to ensure all resources are properly generated. A test publish to Maven local has been performed to ensure all modules generate a separate artifact and a source artifact. I depended on the different Maven artifacts to make sure add-ons still work.

# Leftovers

- [x] clean up unnecessary dependencies
- [x] move repositories to places where they are actually needed
- [x] use the mod in an actual add-on to ensure compatibility
- [x] ~~publish interface injection and access widener data~~
- [x] clean up the POM of the Maven artifacts to avoid publishing unnecessary transitive dependencies
- [x] cherry-pick the commits from the mapping branch
- [x] check why some models are missing (Accelerator Ring, Fluid Pipe, etc)


Rearth:
Hi, just did some more testing, and this is looking good. Huge thanks again for all the work here. I'll create a test tag after merging to see if the maven upload is working as expected.

PR #514, opened 2025-08-09
Fix fluid detection

MUKSC:
# Description

Changes how the mod detects fluids by replacing most `FluidState.isStill()` calls with `!FluidState.isEmpty()`.

Currently, the mod does not treat flowing fluid blocks as fluids, leading to various issues, including:
- The Enderlic Laser keeps shooting at flowing fluids, while it should ignore them, just like still fluids.
- The Pump won't start extracting from a flowing fluid, despite being capable of searching and extrating the source blocks.
- The Destroyer Block attempts to break flowing fluids, while it should ignore them, just like still fluids. It slows down the quarry operations immensly, as water and lava have an unreasonably high hardness value of 100.
- The Block Hole can be easily nullified with water, as it gets stuck keep absorbing flowing water.

The one call in the `PumpBlockEntity::serverTick` was intentionally left unchanged to prevent infinite fluid generation, and a comment was added to clarify that.

# How Has This Been Tested?

I did a side by side testing to ensure the issues described above has been resolved.

- [x] Confirmed that the original issues have been resolved.

# Checklist:

### DIFF
--- a/common/src/main/java/rearth/oritech/block/behavior/LaserArmBlockBehavior.java
+++ b/common/src/main/java/rearth/oritech/block/behavior/LaserArmBlockBehavior.java
@@ -110,7 +110,7 @@ public class LaserArmBlockBehavior {
                 if (world.getTime() % 40 == 0) {    // periodically reset target
                     return false;
                 }
-                if (blockState.isAir() || blockState.getFluidState().isStill()) return false;
+                if (blockState.isAir() || !blockState.getFluidState().isEmpty()) return false;
                 
                 blockState.randomTick((ServerWorld) world, blockPos, world.random);
                 ParticleContent.ACCELERATING.spawn(world, Vec3d.of(blockPos));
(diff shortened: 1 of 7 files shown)


Rearth:
Hi, thanks for the PR. The changes look good to me. However I'm currently in the process of doing some internal mapping changes for Oritech, so if it's okay with you I'll leave the PR open till that's done and then merge it, as it's a lot easier for me this way.

MUKSC:
No problem at all, thanks for the update!

Rearth:
Hi, thanks again for the PR. In the past few weeks I've changed Oritechs build system and migrated the mod to mojmaps. This means that all files have been changed or changed. I went ahead and just copied your changes with the new mappings and implemented them myself. I marked you as co-author of the commit, instead of trying to update the PR, as that's just easier for me. I hope that's okay with you.

PR #515, opened 2025-08-10
QoL changes to small energy storage blocks

MUKSC:
# Description

Some changes to the Portable Energy Storage and Creative Energy Storage to improve user experience.

- Sneaking now inverts their placement direction.
- They are now wrenchable to change their target (output) direction.
  - Using the wrench sets their target direction to the face that was clicked.
  - If the block is already facing the same way as the clicked face, it inverts the direction instead.

# How Has This Been Tested?

I checked the in-game behaviour of both the Portable Energy Storage and Creative Energy Storage to make sure everything is working as intended.

# Checklist:

### DIFF
--- a/common/src/main/java/rearth/oritech/block/blocks/pipes/GenericPipeBlock.java
+++ b/common/src/main/java/rearth/oritech/block/blocks/pipes/GenericPipeBlock.java
@@ -208,7 +208,7 @@ public abstract class GenericPipeBlock extends AbstractPipeBlock implements Wren
     }
     
     @Override
-    public ActionResult onWrenchUse(BlockState state, World world, BlockPos pos, PlayerEntity player, Hand hand) {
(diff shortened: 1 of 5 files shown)


Rearth:
Hi, thanks for the PR. The changes look good to me. However I'm currently in the process of doing some internal mapping changes for Oritech, so if it's okay with you I'll leave the PR open till that's done and then merge it, as it's a lot easier for me this way.

Rearth:
Hi, thanks for the PR again. I did take a closer look at the changes and decided not to merge this for now, sorry. Having the blocks just rotate without any indication on how/why they rotate doesn't fit too well. It's also marked as pipe wrench and not intended to be applicable to other blocks. If we allow its usage on the small storage I know the next request to allow it on the assembler or something will come next.

PR #542, opened 2025-09-16
improve build workflow

rlnt:
# Description

This pull request improves the build workflow.

The triggers were adjusted to only run on this specific branch and only when relevant files are changed to prevent running the build workflow when something like the README is updated.

The matrix configuration was removed because it only uses a single OS anyway.

A concurrency check has been added to cancel a previously running build workflow if a new one is started simultaneously.

The Gradle wrapper validation action has been removed because it constantly fails due to reaching a timeout. This doesn't impact the functionality, though. The wrapper validation is now part of the [setup-gradle](https://github.com/gradle/actions/tree/main/setup-gradle) action.
This is mentioned [here](https://github.com/gradle/actions#the-wrapper-validation-action).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,39 +1,53 @@
-# Automatically build the project and run any configured tests for every push
-# and submitted pull request. This can help catch issues that only occur on
-# certain platforms or Java versions, and provides a first line of defence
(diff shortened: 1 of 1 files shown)


Rearth:
Looks very clean, huge thanks.

PR #556, opened 2025-10-05
Mixin fixin

### LINKED ISSUE #540: Oritech's LivingEntityMixin seems to interfere with other mod (Let Me Despawn) clearing NBT data on item drop (spawnAtLocation in ItemEntity)
### Issue description

So the mod [Let Me Despawn](https://www.curseforge.com/minecraft/mc-mods/let-me-despawn) has a function where items picked up by mobs are tracked and get a `DataComponents.CUSTOM_DATA` tag added ("picked"), so that when those mobs are despawned (main function of the mod, allowing mobs who picked up items to despawn instead of sticking around) those items still get dropped instead of disappearing.

Its library/dependency by the same author, [Almanac Lib](https://www.curseforge.com/minecraft/mc-mods/almanac-lib), seems to be what handles that function where when an item is dropped, it checks for this CompoundTag's existence on the item, and if "picked" is present, it removes it. This seems to be handled by an EntityMixin injecting at `spawnAtLocation` in `net/minecraft/world/item/ItemStack` and `net/minecraft/world/entity/item/ItemEntity`.

That functionality seems to break when Oritech is also present (minimal other mods, just dependencies mainly - when Oritech is disabled, this functionality works again). I've confirmed this by building Oritech but with its `neoforge/src/main/java/rearth/oritech/neoforge/mixin/LivingEntityMixin` present, and that also fixes it just as if Oritech was disabled. I suspect some kind of compatibility issue with how both sides try to do something with `ItemStack` mixing (or something else in that mixin file).

My reproduction for this was a minimal 1.21.1 NeoForge (21.1.208) instance, with the following modlist:
```
almanac-1.21.x-neoforge-1.0.2 (Let Me Despawn dependency)
architectury-13.0.8-neoforge (Oritech dependency)
athena-neoforge-1.21-4.0.2 (Oritech dependency)
geckolib-neoforge-1.21.1-4.7.6 (Oritech dependency)
jei-1.21.1-neoforge-19.22.1.316
kubejs-neoforge-2101.7.1-build.181 (to conveniently enable showing custom NBT data in advanced item tooltips)
letmedespawn-1.21.x-neoforge-1.5.0
oritech-neoforge-0.17.4
owo-lib-neoforge-0.12.15.5-beta.1+1.21 (Oritech dependency)
rhino-2101.2.7-build.77 (KubeJS dependency)
```
Disabling just Oritech in the modlist above (or building it with the LivingEntityMixin in Oritech/Neoforge removed) entirely fixes this behaviour.

Annoyance of this is that many mob drops in various mob farms randomly get this extra NBT data added, which means there'll be 2 versions of a lot of loot - a regular one without this data, and some with this NBT data, and these two won't stack so all these loot items will take up 2x the item slots each.

Some issues (highly likely or for sure) related to this: https://github.com/frikinjay/let-me-despawn/issues/51, https://github.com/AllTheMods/ATM-10/issues/2963, https://github.com/AllTheMods/ATM-10/issues/3079, https://github.com/AllTheMods/ATM-10/issues/3099, https://github.com/AllTheMods/ATM-10/issues/3271, https://github.com/FTBTeam/FTB-Modpack-Issues/issues/7922

### Steps to reproduce

With the above minimal NF instance (or any modpack using these, I'm coming from FTB Evolution but other packs like ATM10 get it too), I can reproduce it simply by doing the following:
1) Set game difficulty to Hard (this allows some zombie spawns to pick up items), set time to night
2) Dig a hole 2 deep for some place to spawn and contain mobs
3) Spawn bunch of zombies in the hole and throw a bunch of rotten flesh to them, hoping there's at least one zombie in the bunch that spawned capable of picking up items
4) If no zombie picked up any rotten flesh, reset (e.g. `/kill @e[type=!player]`), and go back to 3, if at least one picked up some of the thrown rotten flesh, go to 5
5) Kill at least the zombie that picked up a rotten flesh to obtain its drop including its held item, that flesh it picked up
6) Note how this rotten flesh that went thrown to zombie => picked up by zombie => zombie slain => dropped again and picked up by player again, now has a new Custom NBT data called "picked" set to true (1b), visible in F3 + H advanced tooltips and holding Alt (KubeJS added). This means this rotten flesh now won't stack with other rotten flesh that lacks this NBT data.

### Minecraft version

1.21.1 (Latest)

### Platform

Neoforge

### Loader version

21.1.208

### Oritech version

0.17.3 (CF) and 0.17.4 (github build)

### Other relevant versions

_No response_

_No response_

ChiefArug:
# Description

This was originally going to fix only https://github.com/Rearth/Oritech/issues/540 however after seeing the state of Oritech's mixins this is a complete pass over all of Oritech's mixins.

Changes:
- Fixes a bug where the common mixins weren't actually being applied to NeoForge, likely due to #501 
- Investigate the self mixins for extending the AABB's on NeoForge. These were not able to be inlined solely for the reason that `AABB.INFINITE` is a NeoForge patch, however if proper bounds were provided for these I suspect this self mixin could be removed. They have been compacted into one mixin as they all do the same thing.
- Move Elytra layer to platform specific Impls. NeoForge has a patch that causes the mixin to fail on Neo, so we use this patch to add it using the proper Neo API.
- Fixes the absolutely horrendous LivingEntityMixin that is used to teleport drops to the laser when it kills an entity.
- Stop using the fake player username to identify which players are a fake player. That is a valid username that is still available (if someone wants free drop teleportation on older versions of Oritech snag it quick!). I replaced it with a marker interface instead
- The attribute copy mixin was removed and replaced with Architectury's Player Clone event. Also removed the random println.

Fixes #540

# How Has This Been Tested?

- Elytra mixin has been tested on both NeoForge and Fabric with player, zombies and armour stands
- I do not know enough about Oritech to test the attribute copying, but I doubt it needs testing anyway as no functionality was changed.
- I do not know enough about Oritech to test the AABB bounds mixin, but I doubt it needs testing anyway as no functionality was changed.
- I do not know enough about Oritech to test the laser fake player drops. I see no reason for it to break but this should be tested by someone with more knowledge than me.
- #540 has been tested in production and fixes it compared to 0.17.3. Note that you cannot reproduce this issue on 0.18 due to the mixin not applying as mentioned above.

# Checklist:


ChiefArug:
Oops, looks like I forgot to remove the @Debug(export = true) annotation from one of the mixins, that will need removing.

Rearth:
Hi, just did some testing from your branch, and things are looking great. I didn't know you can just define multiple mixin configs, this makes things much better. Huge thanks for the improvements and for dealing with my ugly mixins.

PR #566, opened 2025-10-16
Fix: Restrict soul collection to LivingEntity deaths only

### LINKED ISSUE #563: Boats shouldn't have souls
### Issue description

Destroying boats near an Arcane Catalyst will release souls, allowing players to cheese the mechanic. The docs state that "When a mob or living entity dies, it releases a soul" - I don't think boats should count as living entities?

`BaseSoulCollectionEntity` doesn't appear to do any entity filtering. Perhaps it should filter to only entities that extend `LivingEntity`?

### Steps to reproduce

1. Place boat near Arcane Catalyst
2. Destroy boat
3. Check Arcane Catalyst soul count in GUI

### Minecraft version

1.21.1 (Latest)

### Platform

Neoforge

### Loader version

21.1.206

### Oritech version

0.18.2

### Other relevant versions

_No response_

_No response_

stal111:
# Description

Currently, all entities release souls after death. This means that entities like Boats, Paintings, Armor Stands and such can be used for soul collection, which seems wrong. 
This pr fixes this by making sure the entity extends LivingEntity.

Fixes #563 

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/common/src/main/java/rearth/oritech/block/entity/arcane/BaseSoulCollectionEntity.java
+++ b/common/src/main/java/rearth/oritech/block/entity/arcane/BaseSoulCollectionEntity.java
@@ -4,6 +4,7 @@ import java.util.HashSet;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Holder;
 import net.minecraft.server.level.ServerLevel;
+import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.level.block.entity.BlockEntity;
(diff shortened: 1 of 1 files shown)


Rearth:
Hi, thanks for the PR. However this will break the soul flowers, right?

stal111:
Yeah I think it currently does, forgot those existed. Since the Soul Flowers fire the game event with a null entity I could change it to allow null || instanceof LivingEntity.

Rearth:
Looks good now, thanks

PR #591, opened 2025-11-09
Fix docs

Tctiancan:
Add missing icon for refinery_block, shrinker_block, unstable_container.
Add shrinker.mdx to interaction/_meta.json.
oil_bucket -> still_oil_bucket
Add a blank line before the particle_dist_calc to fix image not display ingame.
Each burst addon increase burst time 4 seconds -> 12 seconds. (240 ticks / 20 = 12 seconds)
https://github.com/Rearth/Oritech/blob/2ecc97e9eb4caf7d45a434de94fe89fd089cb0a5/common/src/main/java/rearth/oritech/init/Config.java#L223

- [x] Feature has been local preview with previewDocs task.


Rearth:
Very nice, huge thanks for adding those. I totally missed them and since I don't use the wiki that often I'd have totally missed it.

PR #593, opened 2025-11-10
add nodules to the config file

Samuel-Chapuis:
# Description
This pull request aim to add an option in the config file to able/disable the ore nodes that spawn on the surface of the map.
For the context, I'm one of the dev in my friend's modded server and we wanted to add Oritech in the new version but nodes was considered as to easy to farm by the comunity so I wanted to disable this feature to be able to use the orithech.

Fixes # (issue)

# How Has This Been Tested?

The changed have been tested on client and server side I havn't seen any issue while launching the game and generating the map with true/false in the config.

IMPORTATNT it have been tested only for neoforge

- [no] Feature has been tested ingame on both neoforge and fabric.

# Checklist:

- [no] My code uses the 'var' keyword where applicable.
- [yes] I have commented my code, particularly in hard-to-understand areas ( but in fact there is not that much complex part )

Additionaly, I would like to thank you for the mod it is really top tiers !

### DIFF
--- a/common/src/main/java/rearth/oritech/init/Config.java
+++ b/common/src/main/java/rearth/oritech/init/Config.java
@@ -74,6 +74,7 @@ public class Config {
     @SectionHeader("worldGeneration")
     public boolean generateOresFabricOnly = true;
     public boolean easyFindFeatures = true;
+    public boolean generateResourceNodes = true;
     
     @SectionHeader("reactor")
     public boolean safeMode = false;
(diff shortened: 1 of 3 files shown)


Rearth:
Hi, this is already covered with the "easyFindFeatures" config option. The featurecontent file is also not applied on neoforge, as neoforge has feature configuration done via json files. In your case I'd also highly recommend just lowered the amount of ores per boulder instead of disabling them, as that makes a relevant part of Oritech (the bedrock resource nodes) basically impossible to find.

PR #595, opened 2025-11-12
Remove @Debug annotation from ElytraLayerMixin

ChiefArug:
# Description
Removed the `@Debug` annotation from ElytraLayerMixin.

I forgot to do this as a part of #556. The annotation is mostly harmless, it just causes mixin to write a class file to the .mixin.out folder when the elytra layer class is loaded.


Rearth:
Nice, thanks for the cleanup PR

PR #706, opened 2026-02-20
Use slot-by-slot insertion to resolve Fluid Tank transfer issue with Item Pipes

### LINKED ISSUE #705: Item Pipes Delete Identical Fluid Tanks During Transfer
### Issue description

When an inventory containing multiple Fluid Tanks with the same fullness (e.g. `n` "Turbofuel Fluid Tanks` each containing 256B) is transferred to another chest via an Item Pipe, only one tank will end up in the target chest. The other tanks disappear during the transfer. (See reproduction steps for more detail).  

For thoroughness, I tried to reproduce this in Neoforge as well. It appears to be Fabric-specific - Fluid Tanks transfer as expected in the Neoforge version. 

mclogs link is attached, however, on a cursory review, there does not seem to be an obvious line in the logs that points to the issue. 

### Steps to reproduce

The simpler reproduction is something like this:

Image

Here is the initial content of the "source" chest on the left: 

Image

When I set the pipe on the source chest to extract into the target chest, I wind up in this state. Notice that, though there were several chests of each "fullness" in the source chest, only one of each exists in the target. The others seem to have simply disappeared.

Image

It is worth noting that this problem is not specific to Portable Tanks, but rather seems to be an issue in any case where an item uses DataComponents.MAX_STACK_SIZE to change from the default stack size. For a simpler reproduction (read: no faffing about with filling Portable Tanks), the issue can be produced with some diamonds given via `/give @s diamond[max_stack_size=1]`. 

### Minecraft version

1.21.1 (Latest)

### Platform

Fabric

### Loader version

0.18.4

### Oritech version

1.0.1

### Other relevant versions

_No response_

https://mclo.gs/D2ps4Yz

epiccoleman:
# Description

This PR attempts to address #705 

As described in the issue, when multiple filled Portable Tanks (or Portable tanks with duplicate fullness values) are transferred into an inventory via an Item Pipe, the source chest will receive only one of the Portable Tanks, with the subsequently transferred Portable Tanks disappearing. 

It looks like Fabric's `Storage.insert()` method does not respect the item's `DataComponents.MAX_STACK_SIZE` when merging into existing slots in vanilla inventories. This may imply that the problem exists for other items as well, though I did not find any other cases during testing. 

The fix adds a slot-by-slot transfer for `SlottedStorage` inventories in the `FabricItemApi` wrapper class, to ensure that the inserted item's `MaxStackSize` is respected. For non `SlottedStorage` inventories, the original behavior is preserved. 

**Full transparency: I used Claude Code to help make this PR.**

Further disclaimer: I know very little about Minecraft mod dev - so I am not sure that this is the correct approach to the bugfix or if this approach might carry other implications. 

Fixes #705 

# How Has This Been Tested?

This was tested in game on a fresh instance of Minecraft 1.21.1 with only the Fabric loader and Oritech 1.0.1 installed. I used the same test setup as described in #705 - two chests next to each other, with a source chest being extracted to the target chest. I filled several tanks of both Turbofuel and Sulfuric Acid and verified that the target chest received each tank as expected. 

It is worth noting that this problem is not specific to Portable Tanks, but rather seems to be an issue in any case where an item uses DataComponents.MAX_STACK_SIZE to change from the default stack size. I also tested the fix with some diamonds given via `/give @s diamond[max_stack_size=1]`, which reproduced the problem behavior in an unmodified Oritech install, and which also behaves appropriately after applying this PR.

### DIFF
--- a/fabric/src/main/java/rearth/oritech/fabric/FabricItemApi.java
+++ b/fabric/src/main/java/rearth/oritech/fabric/FabricItemApi.java
@@ -69,6 +69,36 @@ public class FabricItemApi implements BlockItemApi {
         @Override
         public int insert(ItemStack inserted, boolean simulate) {
             if (inserted.isEmpty()) return 0;
+
+            // Use slot-by-slot insertion to respect item max stack size,
+            // which Fabric's bulk insert may not check for component-based MAX_STACK_SIZE
+            if (storage instanceof SlottedStorage<ItemVariant> slottedStorage) {
+                var variant = ItemVariant.of(inserted);
+                var maxStackSize = inserted.getMaxStackSize();
+                var remaining = inserted.getCount();
+
+                try (var transaction = Transaction.openOuter()) {
+                    for (int i = 0; i < slottedStorage.getSlotCount() && remaining > 0; i++) {
+                        var slot = slottedStorage.getSlot(i);
+                        var currentAmount = (int) slot.getAmount();
+
+                        // Skip slots that are full according to the item's max stack size
+                        if (!slot.isResourceBlank() && currentAmount >= maxStackSize) continue;
+
+                        var canInsert = Math.min(remaining, maxStackSize - (slot.isResourceBlank() ? 0 : currentAmount));
+                        if (canInsert <= 0) continue;
+
+                        var insertedCount = (int) slot.insert(variant, canInsert, transaction);
+                        remaining -= insertedCount;
+                    }
+
+                    if (!simulate)
+                        transaction.commit();
+                    return inserted.getCount() - remaining;
+                }
+            }
+
+            // Fallback for non-slotted storage
             try (var transaction = Transaction.openOuter()) {
                 var insertCount = storage.insert(ItemVariant.of(inserted), inserted.getCount(), transaction);
                 if (!simulate)


epiccoleman:
I was thinking about this a little more during the work day - I wonder if the fix for this issue really should target Fabric, since it does seem to be a bug in their API more so than in Oritech specifically (I'd think that Oritech should be able to expect that Storage.insert won't just disappear items, instead of special casing SlottedStorage which I assume is the most common type). 

I'll take a glance at the Fabric issue tracker and see if there's anything in there that seems related, and report back. Probably still a case to be made for addressing it in Oritech in the interim?

epiccoleman:
OK, further investigation. I think I've found the issue in Fabric itself. 

I think what's happening here is that `InventorySlotWrapper.getCapacity` is using the Fluid Tank item definition's default, unfilled stack size of 64. When a Fluid Tank is filled its maximum stack size is changed via DataComponents, and so Fabric is unaware that the tank can't stack, so it tries to push it into the same slot as the other tank. I would need to spend a bit more time than I have to trace right now to trace `the` whole chain and debug it.    

```
	@Override
	public int getCapacity(ItemVariant variant) {
		// Special case to limit buckets to 1 in furnace fuel inputs.
		if (storage.inventory instanceof AbstractFurnaceBlockEntity && slot == 1 && variant.isOf(Items.BUCKET)) {
			return 1;
		}

		// Special case to limit brewing stand "bottle inputs" to 1.
		if (storage.inventory instanceof BrewingStandBlockEntity && slot < 3) {
			return 1;
		}

		return Math.min(storage.inventory.getMaxCountPerStack(), variant.getItem().getMaxCount());
	}
```
[relevant Fabric source](https://github.com/FabricMC/fabric-api/blob/796c7622f64a322dee91d98d60a4a61a906c55b1/fabric-transfer-api-v1/src/main/java/net/fabricmc/fabric/impl/transfer/item/InventorySlotWrapper.java#L112C1-L125C3)

I've built a tweaked Fabric which changes that last line to: 

```
return Math.min(storage.inventory.getMaxCountPerStack(), variant.toStack().getMaxCount());
```

Which I think will respect the specific item instance's max stack size. We'll see how far I get, I did not expect to be building a new copy of Fabric today :)

epiccoleman:
Alright. Built a fresh copy of Fabric API with my change. That does indeed fix the issue. 

I'll get an issue opened up on the Fabric repo with a PR over there, and link here when that's done. @Rearth - I'll leave this issue open over here for now - guess it's up to you whether you'd want Oritech to fix this in the Fabric wrapper code or wait for a lower-level fix in Fabric.

epiccoleman:
I was able to confirm that this issue can be fixed in Fabric API. I've filed an [issue](https://github.com/FabricMC/fabric-api/issues/5219) and a [PR](https://github.com/FabricMC/fabric-api/pull/5220) over on the Fabric API, which explains the problematic chain of calls in the Fabric's Storage API and proposes a fix.

epiccoleman:
The bugfix associated with this for Fabric has been merged, so this PR is probably no longer needed.

Rearth:
Hi, sorry for the delayed response, I've been busy with some other stuff. Huge thanks for taking care of this issue and getting the improvement fixed upstream in fabric API. This is great to hear. I'll close this PR now as it's no longer needed.

PR #725, opened 2026-03-28
Use tool for loot to enable silk touch / fortune

jshipley:
# Description

When using a chainsaw or prometheum axe to chop down an entire tree, the enchantments on the tool will now be respected when it comes to calculating loot drops.

# How Has This Been Tested?

I have tested chopping down trees with enchanted and unenchanted chainsaws and prometheum axes on Fabric and NeoForge. I get leaves when the axe has Silk Touch as expected.

# Checklist:

- [ X ] My code uses the 'var' keyword where applicable.
- [ X ] I have commented my code, particularly in hard-to-understand areas

There wasn't really any new code, I just replaced the <Level, BlockPos> Tuple with a record that also contains the axe ItemStack, and changed which version of Block.getDrops() gets called.


Rearth:
Hi, thanks for the PR. I'm currently in the process of cleaning up some other stuff but will take a look at this real soon.

Rearth (approved):

Rearth:
Thanks for looking into this, this looks like a pretty clean solution.

PR #755, opened 2026-05-03
Large particle accelerator performance optimization

JuNi4:
# Description

I added extra caching for the particle accelerators.
This was done because I wanted to be able to reach over 1% of light speed.
This needs a particle accelerator that has very long sections (7km for light speed), causing lag at high speeds as the particle checks blocks in unloaded chunks and thus loads them as it is going through them at a rate of hundreds of chunks per second.
So I simply cached the result of the block state queries and it ran much smoother. However, there is still lag at extremely high velocities.

The second change I did was changing the particle speed type from float to long since it never uses decimal values, as far as I could tell, and floats can get stuck at certain numbers when just adding one.
It can now accelerate up to at least 10% of light speed.

# How Has This Been Tested?

- [x] Was tested in a survival world I have been playing in and did not notice any problems.

# Checklist:


Rearth:
Hi, thanks for the PR, it's looking pretty good. I never really intended the particle accelerator to be able to reach fractional light speed, but performance optimizations are always good. I'll do some testing this weekend and then hopefully merge this PR, huge thanks already.

Rearth:
Looks good, huge thanks for the PR

PR #811, opened 2026-08-07
Rename Dubios to Dubious in the back-end

jshipley:
# Description

This fixes old typo, and all of the many things that depend on it.

This changes everything (including identifiers) to use the word "dubious" instead of "dubios".

The "equipmentCharger" id was also changed to "equipmentcharger", because Oritech was refusing to compile for me with upper case letters in the id string.

# How Has This Been Tested?

- [X] Feature has been tested ingame
- [X] Optional additional testing details

Output JSONs checked, and I've confimed that the item still exists, is still getting used in recipes, and is getting the correct translation (in en_us)

# Checklist:

- [N/A] My code uses the 'var' keyword where applicable.
- [N/A] I have commented my code, particularly in hard-to-understand areas

It's pretty much just a string replace, no new comments or variables needed


Rearth:
Hi, thanks for this fix. This is something I should've done far earlier 😄

PR #812, opened 2026-08-07
Adding inter-mod compatibility back into Oritech for NeoForge 26.1.2

jshipley:
# Description

I took a completely different (and much more conservative) approach to compatibility for 26.1.2 than I did for 1.21.1

Some of the main goals this time around:

* Accept / process items from AllTheOres and FTB Materials
* Add AllTheOres & FTB Materials to Ore Boulders and Resource Nodes if they're available
* When outputting items processed with Oritech machines, try to output vanilla items > Oritech items > ATO/FTB items
* Add some basic compatibility with other mods where supporting them specifically makes sense (such as being able to cast molten nickel to oritech:nickel_ingot if ATO and FTB Materials aren't available)
* Try to respect mod identities (much less processing/alloying of Oritech materials in other mods and other mods' materials in Oritech)
* Add some additional fun/useful interactions
* I did **NOT** add any new configuration options, or take any steps to make Oritech produce ATO / FTB Materials items instead of Oritech items. This might be a possible future improvement to make unification easier for pack creators.

This is a semi large change, so I'll try to summarize the main components:

* Added a new runCompatData task in build.gradle
    - added new compat sourceSet, dependency types (compatImplementation, compatCompileOnly, compatRuntimeOnly) that will add dependencies to compat without affecting the main build
    - compat code has access to Oritech code, but Oritech doesn't have access to compat code
    - src/compat/generated/resources will be included in Oritech build
    - Oritech data will be generated as part of the runCompatData task, but compat data will not be generated as part of a runData task
    
* Changed ResourceNodeFeature configuration (**breaking change** for any data packs that have customized this)
    - nodeOres and boulderOres block lists have become weighted block lists, with optional weight (default 1), required (default true), and requiresAny (default empty list) fields
    - "requiresAny" takes a list of block IDs, and the weighted block will be ignored if none of the listed blocks are registered
    - Technically, repeated list elements are OK but unnecessary. You can just have {"block": "minecraft:deepslate", "weight": 30} instead of repeating "minecraft:deepslate" 30 times.
    - Generic resource ores were added (basically recolored gold ore with a deepslate background) that will appear in bedrock below boulders that can have the deepslate version of those ores
    - AllTheOres and FTB Materials are configured by default now, and will generate in the ore boulders and resource nodes if those mods are loaded

* Recipe builders updated a little to allow for a "prefix" in recipe IDs
    - All existing datagen recipes were unaffected
    - Compat recipe IDs now all start with "compat/<modname>" (eg "compat/ae2/pulverizing/certus_dust" instead of "pulverizing/compat/ae2/certus_dust"

# How Has This Been Tested?

- [X] Feature has been tested ingame
- [X] Optional additional testing details

For the world gen changes, I have tested with a combination of ATO and FTB Materials being loaded and unloaded. Their ores are included in the ore boulders when the mods are loaded, and not included if they're not loaded.

I've also reviewed the JSON files for the other recipes and made sure they looked good in JEI.

# Checklist:

- [X] My code uses the 'var' keyword where applicable.
- [X] I have commented my code, particularly in hard-to-understand areas

Please let me know if there are any compat features that existed in 1.21.1 that I skipped that you would prefer to include in 26.1.2


Rearth:
Hi, thanks for the PR, and sorry for the late review. This approach generally sounds good as it doesn't add too much. Breaking changes are okay in the current state of the mod, so this isn't a problem. Not sure how I feel about always registering the other mods materials as bedrock ores. Would they still show up in the create tab / JEI even if the other mods aren't loaded?

jshipley:
Yes, it would always register those generic resource nodes whether there are ores for them or not, and they would show up in JEI even if they're irrelevant.

One reason why I left them always registered even if they're not used is that I didn't want to try to guess which mods people might want to use them with. Certainly ATO and FTB Materials, but there's also FTB Industrial Contraptions, Mekanism (if that gets updated), Occultism, Bronze, Energized Power, LTX Industries, and many other mods which might add some of those ores that could be configured by a modpack to use the resource ores.

That might be a good place for a configuration switch... Maybe load the relevant blocks if ATO or FTB Materials is loaded, or if they're in a list of resource nodes to register in the config? So if you want to enable a resource node for Occultism silver without ATO or FTB Materials loaded you'd just have to enable it in the config and add the bedrock extractor recipe.

PR #823, opened 2026-08-17
Optimize energy pipe network distribution

FireController1847:
# Description

Reworks energy pipe transfer to run once per network per tick with shared runtime state, rotating source/target fairness, and topology revision invalidation instead of per-interface target shuffling. Adds `AggregatingEnergyStorage` to extract across multiple interface buffers without buffering, prevents self-feedback loops, tracks network revision cleanup in `PipeNetworkData`, and clears runtime caches on server stop.

# How Has This Been Tested?

Tested on NeoForge & Fabric. Functions as intended. Significant performance improvements when many are connected like on a very large reactor output.

# Checklist:

### DIFF
--- a/common/src/main/java/rearth/oritech/Oritech.java
+++ b/common/src/main/java/rearth/oritech/Oritech.java
@@ -23,6 +23,7 @@ import rearth.oritech.block.entity.augmenter.PlayerAugments;
 import rearth.oritech.block.entity.augmenter.api.Augment;
 import rearth.oritech.block.entity.interaction.PowerPoleEntity;
 import rearth.oritech.block.entity.pipes.GenericPipeInterfaceEntity;
+import rearth.oritech.block.entity.pipes.EnergyPipeInterfaceEntity;
 import rearth.oritech.client.init.ModScreens;
 import rearth.oritech.client.init.ParticleContent;
 import rearth.oritech.init.*;
@@ -59,6 +60,7 @@ public final class Oritech {
         
         // for pipe data
         LifecycleEvent.SERVER_STARTED.register(Oritech::onServerStarted);
+        LifecycleEvent.SERVER_STOPPED.register(server -> EnergyPipeInterfaceEntity.clearRuntimeData());
         
         // for augment data
         LifecycleEvent.SERVER_STARTED.register(server -> PlayerAugments.loadAllAugments(server.getRecipeManager()));
--- a/common/src/main/java/rearth/oritech/api/energy/containers/AggregatingEnergyStorage.java
+++ b/common/src/main/java/rearth/oritech/api/energy/containers/AggregatingEnergyStorage.java
@@ -1,0 +1,115 @@
+package rearth.oritech.api.energy.containers;
+
+import rearth.oritech.api.energy.EnergyApi;
+
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * A read-only view over multiple energy storages.
+ *
+ * <p>Extraction is spread across the backing storages without moving their energy into an
+ * intermediate buffer. Each backing storage keeps its own extraction limit, so the aggregate's
+ * throughput is the sum of the throughput available from its members.</p>
+ */
+public class AggregatingEnergyStorage extends EnergyApi.EnergyStorage {
+
+    private final List<? extends EnergyApi.EnergyStorage> storages;
+    private final long[] remainingExtractable;
+    private final Set<EnergyApi.EnergyStorage> changedStorages = Collections.newSetFromMap(new IdentityHashMap<>());
+    private long amount;
+    private long extractableAmount;
+    private long capacity;
+    private int extractionIndex;
+
+    public AggregatingEnergyStorage(List<? extends EnergyApi.EnergyStorage> storages) {
+        this.storages = List.copyOf(storages);
+        this.remainingExtractable = new long[storages.size()];
+
+        for (var i = 0; i < storages.size(); i++) {
+            var storage = storages.get(i);
+            amount = saturatedAdd(amount, storage.getAmount());
+            capacity = saturatedAdd(capacity, storage.getCapacity());
+
+            if (!storage.supportsExtraction()) continue;
+            remainingExtractable[i] = storage.extract(Long.MAX_VALUE, true);
+            extractableAmount = saturatedAdd(extractableAmount, remainingExtractable[i]);
+        }
+    }
+
+    @Override
+    public boolean supportsInsertion() {
+        return false;
+    }
+
+    @Override
+    public long insert(long maxAmount, boolean simulate) {
+        return 0;
+    }
+
+    @Override
+    public long extract(long maxAmount, boolean simulate) {
+        if (maxAmount <= 0) return 0;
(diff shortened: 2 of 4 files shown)


Rearth:
That seems like a whole lot of extra complexity and processing on the server for very little benefit. What exactly is the problem you're trying to solve here?

FireController1847:
> That seems like a whole lot of extra complexity and processing on the server for very little benefit. What exactly is the problem you're trying to solve here?

I can get pictures later, but as it currently stands there's a very major issue with superconductors where if you connect many of them to a reactor, it will PLUMMET the server's TPS when energy is being drawn from them. Assuming you have a 28x28 reactor, you'll need at LEAST 2,352 energy ports if you're not allowed to edit the config to pull the power needed from it, and that doesn't even really capture all of it (e.g. a reactor producing 42 MFE/t would require a minimum of 2,800 energy ports to pull it all).

I discovered this bug after building two of these reactors, then running 6 sets of 4 superconductors down the sides of it, resulting in a combined total of 12 lines of superconductors running into my energy storage.

Essentially, what happens is when energy begins being drawn from the network, each conductor is ticked individually-but it results in each conductor having to consider the other conductors in the network, resulting in a HUGE performance hit for relatively minor energy draw. It should be quite easily reproducible with a relatively large reactor design.

The purpose of this PR is to resolve the issue of ticking every individual superconductor and it's internal buffer by combining shared networks into one combined buffer, which optimizes throughout and prevents the need to tick each superconductor every tick to pull energy when it's simply transporting from one energy source (the reactor) to another (energy storage). The benefit is that the per-tick performance cost of large superconductor networks is significantly reduced.

For reference, in our 2,800 superconductors example, that ticks 2,800 entities per tick of energy storage transfer, performs 2,800 updates of each entity and ultimately increases the tick time significantly when you have a reactor this size. Given my goal of having eight of these, at 22,400 superconductors my server would grind to a halt.

With this PR, the 22,400 ticks can be reduced to simply 12 per reactor, bringing the total per-tick cost down to 96 instead of 22,400. It just essentially optimizes the big-O of the energy transfer performance for superconductors when scaling upwards. Obviously the drawback is an additional interface, but I strongly believe the gains of this approach significantly outweigh the costs.

Please let me know if you'd like pictures of my setup. I spent about two days debugging the issue between multiple sets of mods and can very reliably reproduce the issue to only exist with Oritech superconductors (rather the energy entity), and it only occurs when power is actively being drawn from them.

Rearth:
Hm, thanks for the detailed response. I'm still not a fan of adding this much complexity that I didn't create into the mod for such a niche problem, since I'm the one maintaining it in the long run.
Also some of the information sounds like either a bug or misunderstand of the energy handling. All oritech machines and pipes operate push-based for energy. So none of the superconductors try to pull any energy. Only the ones that have energy pushed to it from the reactor outputs try to push it to network targets. And only the ones connecting to other machines/energy containers are block entitities. Other superconductors don't count as network targets, so the scaling is only based on the network inputs and outputs. Yes a small optimization  can be made as multiple pipes could try to insert into the same targets. However marging that and handling all the extra cases that come with that sounds like a nightmare to handle, especially for larger / super long pipe networks). 
Also the math on the number of energy ports needed sounds off with default configs. To get to the 2400 ports needed with default configs you'd need an absurdly large reactor, if at all possible. And if the pack changes configs to increase rf generation, then of course you'd need to also update the output rate.

FireController1847:
I'd absolutely be down to try and collaborate to find a better solution if you believe there is one. At the moment regardless, the reproducible behavior is the ticking of Oritech superconductors drops my server's TPS from 20 to 11 whilst energy is being drawn.

It is my understanding that each energy port only allows an output of 15,000 FE/t. If my reactor makes 42,000,000 FE/t, 42,000,000 / 15,000 = 2,800 ports required. Unless I'm mixing up my math there, it's very possible I could be misunderstanding something.

I will be completely transparent I did use Codex's premium plan to help me derive this PR, but all of the code was individually reviewed and tested by myself. Is there any more information I could provide to try and explain the issue further? At the end of the day, as it currently stands my reactor is just, not usable because of the lag which I can reliably trace back to Superconductors, from Spark reports to isolation testing.

Rearth:
But a spark report of a save that has the issue would be interesting. Any chance you have one with a problematic world? There for sure is a lot of optimization potential, but I also want to end up with code that I actually understand and can maintain. Supporting 2000+ active pipe entities for a nuclear reactor is neither a reactor use case nor a pipe usecase I really want to invest a lot of time into.

FireController1847:
Of course. I'm with you there. Finding that balance is very important. I'll try and get you one later today!

Rearth:
With default configs the reactor should output 25000 RF/t IIRC. Also with default configs you really shouldn't easily get a reactor up to 42M RF/t. Are you playing in a public pack?

Rearth:
> Of course. I'm with you there. Finding that balance is very important. I'll try and get you one later today!

Great, thank you

FireController1847:
> With default configs the reactor should output 25000 RF/t IIRC. Also with default configs you really shouldn't easily get a reactor up to 42M RF/t. Are you playing in a public pack?

Yes, FTB Evolution :)

Rearth:
Hm, just downloaded ftb evolution. They have heavily modified their configs. The amount of RF generated is increased by 150x. The output per reactor port is also massively increased to 50 000 000 RF/t. So you should be fine with one or 2 ports. Not sure what version of the pack you're on, but in the current one thats the case. So for most reactors you'll only need 1 or 2 ports. That should solve your problems.
However I'd still be interested in a spark report with more ports/pipes to further optimize things.

FireController1847:
Yes, absolutely. I apologize for my initial combativeness. As it turns out, the GuideMe doesn't reflect the config value and still says 25,000 RF/t LOL (and yes, it was 25k, not 15k; I was misremembering). Man, I wish I knew that. I was a bit confused about whether I could transfer it all, but I assumed it was sufficient for the generation rate. Now what the heck do I do with 22,000 crafted superconductors? LOL

However, I'd still love to help try and find a solution to this! I do love the look of the current setup, to be fair. And it was a fun extra challenge. So, in terms of gathering some actual evidence, here's a screenshot of my current reactor design:

image
image

Here's the profile prior to the reactor running. My TPS isn't great given I just restarted the server so it's still stabilizing, but I'm pretty sure the issue can still be seen in the difference between the two profiles.

[REACTOR_NOT_RUNNING_profile-2026-08-18_09.17.43.sparkprofile.zip](https://github.com/user-attachments/files/31185783/REACTOR_NOT_RUNNING_profile-2026-08-18_09.17.43.sparkprofile.zip)

And here's the profile following when I started the reactor. No other variables were changed–TPS dropped from the measly 15 I had to 10, and has stayed there until all the conductor's buffers have disharged, when it returns back to 15-16.

EDIT: I accidentally uploaded the same profile twice. Here's the fixed one.

[REACTOR_RUNNING_profile-2026-08-18_09.18.33.sparkprofile.zip](https://github.com/user-attachments/files/31186023/REACTOR_RUNNING_profile-2026-08-18_09.18.33.sparkprofile.zip)

Rearth:
Hi, thanks for the detailed reports. I have some ideas for optimizations but think that this PR is a bit of an overkill for this problem, so I'll close this for now. I might release an update soon that makes some smaller optimizations here.

PR #826, opened 2026-08-19
Improved entity management for Dubious Container

Hiiragi283:
# Description

Improved entity management for Dubious Container by using entity tags instead of direct value list.
With this change, players and addon developers can freely customize which entities can be captured via tags.

In addition, I found an potential issue which `ComponentContent.ENERGY` accepts negative energy value.
So I just replaced `Codec.INT` with `ExtraCodecs.NON_NEGATIVE_INT` to resolve this small problem.

# How Has This Been Tested?

# Checklist:

- [N/A] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.2.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.6.1-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME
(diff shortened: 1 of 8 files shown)


Rearth:
Hi, thanks for this PR. This looks mostly good, but why is the entity type tag not used directly in the item code, and is instead put into a data component that is then loaded in the item code? This sounds a lot of extra unneeded complexity, no? Or am I missing something here?

jshipley:
If there's a data component, I'd want to be able to use something like oritech:dubious_container[oritech:dubious_container_target=["minecraft:zombie","minecraft:enderman"]] or even just give it an arbitrary tag during stack creation (eg /give)

Would that work with this change?

Hiiragi283:
> Hi, thanks for this PR. This looks mostly good, but why is the entity type tag not used directly in the item code, and is instead put into a data component that is then loaded in the item code? This sounds a lot of extra unneeded complexity, no? Or am I missing something here?

I used HolderSet because I designed it like other vanilla components.  however, since there is a specific item class, as you mentioned, it might be fine to simply reference entity tag directly.

